PR #11, opened 2020-07-12
Add colab demo description

dreamerlin:


innerlee:
?

dreamerlin:
> ?

Add some description in README.md，and update the the colab link to the master one. (Other wise, it will direct to the wrong link for my fork branch, which has been deleted.)

innerlee:
why change the badges

dreamerlin:
> why change the badges

Ops, It should be a build badge. And I think it is still necessary to indicate the docs site since the badge is Easily overlooked.

PR #14, opened 2020-07-12
Extract rgb frames without denseflow

dreamerlin:
1. Add `--use-opencv` args in `build_rawframes.py` to extract rgb rawframes by OpenCV instead of denseflow
2. Rename sth to MMACTION2

### DIFF
--- a/tools/data/build_rawframes.py
+++ b/tools/data/build_rawframes.py
@@ -7,6 +7,7 @@ import warnings
 from multiprocessing import Pool
 
 import mmcv
+import numpy as np
 
 
 def extract_frame(vid_item, dev_id=0):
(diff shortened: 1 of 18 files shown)


innerlee:
It will be good to support resize in the opencv version of frame extraction.

The benefit of using denseflow is to make sure that the number of flows and the numbers of frames do match exactly.

tools/data/build_rawframes.py
```diff
@@ -29,10 +30,28 @@ def extract_frame(vid_item, dev_id=0):
 
     if task == 'rgb':
         if args.use_opencv:
+            # Not like using denseflow,
+            # Use OpenCV will not make a sub directory with the video name
+            video_name = vid_path.split('.')[0].split('/')[-1]
```
innerlee, line 35: use os.path.basename or os.path.splitext to manipulate paths. avoid string operations when possible

innerlee:
note to future: separate different changes into different prs. this pr is a bad example, which mixes name changes and function changes.

dreamerlin:
> note to future: separate different changes into different prs. this pr is a bad example, which mixes name changes and function changes.

You got that right

tools/data/build_rawframes.py
```diff
@@ -19,21 +23,50 @@ def extract_frame(vid_item, dev_id=0):
     """
     full_path, vid_path, vid_id, method, task = vid_item
     if ('/' in vid_path):
-        vid_name = vid_path.split('.')[0].split('/')[0]
-        out_full_path = osp.join(args.out_dir, vid_name)
+        act_name = osp.splitext(vid_path)[0].split('/')[0]
```
innerlee, line 26: why there is still `split` present?

tools/data/build_rawframes.py
```diff
@@ -19,21 +23,50 @@ def extract_frame(vid_item, dev_id=0):
     """
     full_path, vid_path, vid_id, method, task = vid_item
     if ('/' in vid_path):
-        vid_name = vid_path.split('.')[0].split('/')[0]
-        out_full_path = osp.join(args.out_dir, vid_name)
+        act_name = osp.split(osp.splitext(vid_path)[0])[0]
```
innerlee, line 26: https://www.geeksforgeeks.org/python-os-path-basename-method/

PR #17, opened 2020-07-13
Add explanation for using denseflow to extract both rgb and flow

dreamerlin:


docs/data_preparation.md
```diff
@@ -41,14 +41,15 @@ Note that the videos should be arranged in either
 
 ### Extract frames
 
-To extract frames and optical flow, [dense_flow](https://github.com/open-mmlab/denseflow) is needed.
+To extract both frames and optical flow, [dense_flow](https://github.com/open-mmlab/denseflow) is needed,
+since the benefit of using denseflow is to make sure that the number of flows and the number of frames do math exactly.
```
innerlee, line 45: no need to do math

dreamerlin: Ops

docs/data_preparation.md
```diff
@@ -41,8 +41,8 @@ Note that the videos should be arranged in either
 
 ### Extract frames
 
-To extract both frames and optical flow, [dense_flow](https://github.com/open-mmlab/denseflow) is needed,
-since the benefit of using denseflow is to make sure that the number of flows and the number of frames do math exactly.
+To extract both frames and optical flow, [denseflow](https://github.com/open-mmlab/denseflow) is needed,
+since the benefit of using denseflow is to make sure that the number of flows and the number of frames do match exactly.
```
innerlee, line 45:
```
To extract both frames and optical flow, you can use the tool [denseflow](https://github.com/open-mmlab/denseflow) we wrote.
Since different frame extraction tools produce different number of frame,
it is beneficial to use the same tool to do both frame extraction and the flow computation, to avoid mismatching of frame counts.
```

innerlee:
looks good.

innerlee:
interesting, the auto-merging is a lie

PR #21, opened 2020-07-14
specify a start index for sampling frames

dreamerlin:
Specify a start index for sampling frames in consideration of different filenames format.


innerlee:
did u test the start_index=0 case?

PR #25, opened 2020-07-15
Add fp16 support in action recognition

dreamerlin:
* Use fp16 utils from `mmcv`
* Add fp16 config for TSN
* Add `self.fp16_enable` attribute and `@auto_fp16` tag in `BaseRecognizer` to support fp16 training.
* Ran fp16 setting and normal setting. they got the same accuracy


mmaction/models/recognizers/base.py
```diff
@@ -6,6 +6,7 @@
 import torch.nn as nn
 import torch.nn.functional as F
 
+from ...core import auto_fp16
```
hellock, line 9: https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/fp16_utils.py

tools/test.py
```diff
@@ -7,6 +7,7 @@
 from mmcv.runner import get_dist_info, init_dist, load_checkpoint
 
 from mmaction.apis import multi_gpu_test, single_gpu_test
+from mmaction.core import wrap_fp16_model
```
hellock, line 10: mmaction or mmcv?

hellock (approved):

PR #26, opened 2020-07-15
Reorganize Unittest structure

dreamerlin:
* Migrate unittest files related with data parts into `tests/test_data/` directory.
* Migrate unittest files related with model parts into `tests/test_models/` directory.
* Leave others in `tests/` directory.


innerlee:
why?

dreamerlin:
> why?

Just to make the file structure clearer. I think it is good to do it since we already have 20+ unittest files.

innerlee:
20 is not a big number

dreamerlin:
Don’t you think these files are messy? : )

PR #27, opened 2020-07-15
Add ColorJitter augmentation

dreamerlin:
This pr adds ColorJitter augmentation which includes distorting the brightness, contrast, saturation and hue of images, adding PCA bias to images  . It refers from [mxnet](https://gluon-cv.mxnet.io/_modules/gluoncv/data/transforms/experimental/image.html) and [gluon](https://mxnet.apache.org/api/python/docs/_modules/mxnet/image/image.html#LightingAug).

And I found [TPN repo](https://github.com/decisionforce/TPN/blob/master/mmaction/datasets/transforms.py#L33-L118) also used those for reference but didn't convert the image from BGR to RGB, which may be not correct.

Our codebase the RGB images as input, same with mxnet and gluon.

### DIFF
--- a/mmaction/datasets/pipelines/augmentations.py
+++ b/mmaction/datasets/pipelines/augmentations.py
@@ -697,6 +697,116 @@ class Normalize(object):
 
 
 @PIPELINES.register_module()
+class ColorJitter(object):
+    """
+
+    """
+
+    def __init__(self,
+                 color_space_aug=False,
+                 alphastd=0.1,
+                 eigval=None,
+                 eigvec=None):
+        if eigval is None:
+            # note that the data range should be [0, 255]
+            self.eigval = np.array([55.46, 4.794, 1.148])
+        else:
+            self.eigval = eigval
+
+        if eigvec is None:
+            self.eigvec = np.array([[-0.5675, 0.7192, 0.4009],
+                                    [-0.5808, -0.0045, -0.8140],
+                                    [-0.5836, -0.6948, 0.4203]])
+        else:
+            self.eigvec = eigvec
+
+        self.alphastd = alphastd
+        self.color_space_aug = color_space_aug
+
+    @staticmethod
+    def brightnetss(imgs, delta):
+        if random.uniform(0, 1) > 0.5:
+            delta = np.array(delta).astype(np.float32)
+            imgs = imgs + delta
+        return imgs
+
+    @staticmethod
+    def contrast(imgs, alpha):
+        if random.uniform(0, 1) > 0.5:
+            alpha = np.array(alpha).astype(np.float32)
+            imgs = imgs * alpha
+        return imgs
+
+    @staticmethod
+    def saturation(imgs, alpha):
+        if random.uniform(0, 1) > 0.5:
+            gray = imgs * np.array([0.114, 0.587, 0.299]).astype(np.float32)
+            gray = np.sum(gray, 2, keepdims=True)
+            gray *= (1.0 - alpha)
+            imgs = imgs * alpha
+            imgs = imgs + gray
+        return imgs
+
+    @staticmethod
+    def hue(img, alpha):
+        if random.uniform(0, 1) > 0.5:
+            u = np.cos(alpha * np.pi)
+            w = np.sin(alpha * np.pi)
+            bt = np.array([[1.0, 0.0, 0.0], [0.0, u, -w], [0.0, w, u]])
+            tyiq = np.array([[0.299, 0.587, 0.114], [0.596, -0.274, -0.321],
+                             [0.211, -0.523, 0.311]])
+            ityiq = np.array([[1.0, 0.956, 0.621], [1.0, -0.272, -0.647],
+                              [1.0, -1.107, 1.705]])
+            t = np.dot(np.dot(ityiq, bt), tyiq).T
+            t = np.array(t).astype(np.float32)
+            t = np.flip(t)
+            img = np.dot(img, t)
+        return img
+
+    def __call__(self, results):
+        imgs = results['imgs']
+        out = []
+        if self.color_space_aug:
+            bright_delta = np.random.uniform(-32, 32)
+            contrast_alpha = np.random.uniform(0.6, 1.4)
+            saturation_alpha = np.random.uniform(0.6, 1.4)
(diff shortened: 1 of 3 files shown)


mmaction/datasets/pipelines/augmentations.py
```diff
@@ -697,6 +697,141 @@ def __repr__(self):
...
+        else:
+            self.eig_vec = eig_vec
+
+        self.alpha_std = alpha_std
+        self.color_space_aug = color_space_aug
+
+    @staticmethod
+    def brightnetss(imgs, delta):
+        """Brightness distortion."""
+        if random.uniform(0, 1) > 0.5:
```
innerlee, line 750: use np.random

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -697,6 +697,141 @@ def __repr__(self):
...
+            gray = imgs * np.array([0.114, 0.587, 0.299]).astype(np.float32)
+            gray = np.sum(gray, 2, keepdims=True)
+            gray *= (1.0 - alpha)
+            imgs = imgs * alpha
+            imgs = imgs + gray
+        return imgs
+
+    @staticmethod
+    def hue(img, alpha):
+        """Hue distortion"""
```
innerlee, line 776: what does alpha mean

innerlee:
We may add reference to the referenced code in the docstring

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -697,6 +697,141 @@ def __repr__(self):
...
+    def __call__(self, results):
+        imgs = results['imgs']
+        out = []
+        if self.color_space_aug:
+            bright_delta = np.random.uniform(-32, 32)
+            contrast_alpha = np.random.uniform(0.6, 1.4)
+            saturation_alpha = np.random.uniform(0.6, 1.4)
+            hue_alpha = random.uniform(-18, 18)
+            for img in imgs:
+                img = self.brightnetss(img, delta=bright_delta)
```
innerlee, line 801: brightness

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -697,6 +697,141 @@ def __repr__(self):
...
+
+    def __call__(self, results):
+        imgs = results['imgs']
+        out = []
+        if self.color_space_aug:
+            bright_delta = np.random.uniform(-32, 32)
+            contrast_alpha = np.random.uniform(0.6, 1.4)
+            saturation_alpha = np.random.uniform(0.6, 1.4)
+            hue_alpha = random.uniform(-18, 18)
+            for img in imgs:
```
innerlee, line 800: so images in the same clip does not share the same jittering?

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -697,6 +697,141 @@ def __repr__(self):
...
+                    img = self.saturation(img, alpha=saturation_alpha)
+                    img = self.hue(img, alpha=hue_alpha)
+                    img = self.contrast(img, alpha=contrast_alpha)
+                out.append(img)
+        else:
+            out = imgs
+        alpha = np.random.normal(0, self.alpha_std, size=(3, ))
+        rgb = np.array(np.dot(self.eig_vec * alpha,
+                              self.eig_val)).astype(np.float32)
+        rgb = np.expand_dims(np.expand_dims(rgb, 0), 0)
```
innerlee, line 816: is this `rgb[None, None, ...]`?

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -697,6 +697,141 @@ def __repr__(self):
...
+
+        results['imgs'] = [img + rgb for img in out]
```
innerlee, line 818: what dtype is the imgs? i guess it is still uint8? will it be transformed into float by accident?

innerlee: and why this rgb noise is shared in the image list?

mmaction/datasets/pipelines/augmentations.py
dreamerlin: Why?

innerlee: consistency

dreamerlin: Consistency is good.

mmaction/datasets/pipelines/augmentations.py
dreamerlin: Right

mmaction/datasets/pipelines/augmentations.py
dreamerlin: It adds PCA bias to images, like [this](https://mxnet.apache.org/api/python/docs/_modules/mxnet/image/image.html#CreateAugmenter)

mmaction/datasets/pipelines/augmentations.py
dreamerlin: to control the rotation degree for converting. Add docstring for it

mmaction/datasets/pipelines/augmentations.py
dreamerlin: It is more reasonable that images in the same clip share the same jittering. I add a `jitter_coin` variable to ensure that.

innerlee:
I suggest the following steps:

1. implement the original code, and name the pipeline as `ColorJitter_TPN`
2. implement a cleaner version `ColorJitter`, based on our common sense
3. benchmark the two versions. if accuracy got under `ColorJitter` is not worse than `ColorJitter_TPN`, then delete `ColorJitter_TPN` and merge.

What do you think?

dreamerlin:
> I suggest the following steps:
> 
> 1. implement the original code, and name the pipeline as `ColorJitter_TPN`
> 2. implement a cleaner version `ColorJitter`, based on our common sense
> 3. benchmark the two versions. if accuracy got under `ColorJitter` is not worse than `ColorJitter_TPN`, then delete `ColorJitter_TPN` and merge.
> 
> What do you think?

How about merging it first, since it is more convenient to do benchmark.

innerlee:
no, i do not trust this code, yet

dreamerlin:
Ok, I will benchmark them when I implement TPN

PR #28, opened 2020-07-16
rebenchmark for Pyslowfast comparison

SuX97:
Rebenchmarked on 32G v100 servers.


innerlee:
Could you explain the `re-`? What is the difference in terms of benchmark setting?

SuX97:
> Could you explain the `re-`? What is the difference in terms of benchmark setting?

My part in benchmark.md(slowfast, slowonly, r2+1d, i3d) is conducted on 16G v100 rather than 32G v100, for fair comparison it is redone now.

innerlee:
could you rebase on master?

innerlee:
benchmark in README might need to be updated also

innerlee:
kindly ping

Issue #38, opened 2020-07-20
TSM temporal_pool=True bug


irvingzhang0512:
Thanks for your awesome codabase.
I'm trying to train TSM with `temporal_pool=True`(add `temporal_pool=True` in both `TSMHead` & `ResNetTSM` ) but get some errors.
After some debugging, i think `ResNetTSM` forget to do **actual temporal pool** between layer1 and layer2. 
which means, feature map shape before layer2 should be `N * num_segments/2, C, H, W` instead of `N * num_segments, C, H, W`
In [original TSM codabase](https://github.com/mit-han-lab/temporal-shift-module/blob/master/ops/temporal_shift.py#L92), when `temporal_pool=True`, there is a `max_pool3d` to do **actual temporal pool** before layer2, which is missing in mmaction2.

PR #40, opened 2020-07-20
Support temporal pool in tsm

dreamerlin:
Fix [this issue](https://github.com/open-mmlab/mmaction2/issues/38). 
Since [original repo](https://github.com/mit-han-lab/temporal-shift-module) doesn't use this function but still provides this function in the codes, we supplement it to make temporal pooling between layer1 and layer2, using a 3D max pooling layer and provide the example config file.

### DIFF
--- a/configs/recognition/tsm/tsm_temporal_pool_r50_1x1x8_50e_kinetics400_rgb.py
+++ b/configs/recognition/tsm/tsm_temporal_pool_r50_1x1x8_50e_kinetics400_rgb.py
@@ -1,0 +1,128 @@
+# model settings
+model = dict(
+    type='Recognizer2D',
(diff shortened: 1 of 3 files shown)


irvingzhang0512:
thanks for your fix.
FYI, TSM original codebase use `make_temporal_pool` in [main.py](https://github.com/mit-han-lab/temporal-shift-module/blob/fa8d074d23691686c244dfc2de25c309ba3f284b/main.py#L126)

dreamerlin:
Hi, @irvingzhang0512 ! Thank you for your feedback!

I mean the `make_temporal_pool` is not used in their provided training scripts, so we just ignored it before.

Thanks for your reminding.

PR #41, opened 2020-07-20
Add NonLocal module in TSM and I3D

dreamerlin:


innerlee:
will the trained models be added to modelzoo?

dreamerlin:
> will the trained models be added to modelzoo?

Yes, the trained models will be added to the modelzoo when all the settings are ready.

PR #42, opened 2020-07-21
remove benchmark

SuX97:
remove benchmark for polishing.


dreamerlin:
It seems that this pr is not based on the latest master.

PR #48, opened 2020-07-23
Add feature `with_offset` for rawframe dataset

kennymckormick:
add feature 'with offset' for rawframe dataset, need this feature to train recognition models on untrimmed datasets

### DIFF
--- a/mmaction/datasets/pipelines/loading.py
+++ b/mmaction/datasets/pipelines/loading.py
@@ -591,9 +591,7 @@ class FrameSelector(object):
         if results['frame_inds'].ndim != 1:
             results['frame_inds'] = np.squeeze(results['frame_inds'])
 
-        offset = 0
-        if 'offset' in results:
-            offset = results['offset']
+        offset = results.get('offset', 0)
 
         for frame_idx in results['frame_inds']:
             frame_idx += offset
--- a/mmaction/datasets/pipelines/loading.py
+++ b/mmaction/datasets/pipelines/loading.py
@@ -591,7 +591,12 @@ class FrameSelector(object):
         if results['frame_inds'].ndim != 1:
             results['frame_inds'] = np.squeeze(results['frame_inds'])
 
+        offset = 0
+        if 'offset' in results:
+            offset = results['offset']
+
         for frame_idx in results['frame_inds']:
+            frame_idx += offset
             if modality == 'RGB':
                 filepath = osp.join(directory, filename_tmpl.format(frame_idx))
                 img_bytes = self.file_client.get(filepath)
--- a/mmaction/datasets/rawframe_dataset.py
+++ b/mmaction/datasets/rawframe_dataset.py
@@ -99,18 +99,22 @@ class RawframeDataset(BaseDataset):
                 line_split = line.strip().split()
                 video_info = {}
                 idx = 0
+                # idx for frame_dir
                 frame_dir = line_split[idx]
                 if self.data_prefix is not None:
                     frame_dir = osp.join(self.data_prefix, frame_dir)
                 video_info['frame_dir'] = frame_dir
                 idx += 1
                 if self.with_offset:
+                    # idx for offset and total_frames
                     video_info['offset'] = int(line_split[idx])
                     video_info['total_frames'] = int(line_split[idx + 1])
                     idx += 2
                 else:
+                    # idx for total_frames
                     video_info['total_frames'] = int(line_split[idx])
                     idx += 1
+                # idx for label[s]
                 label = [int(x) for x in line_split[idx:]]
                 assert len(label), f'missing label in line: {line}'
                 if self.multi_class:
--- a/mmaction/datasets/rawframe_dataset.py
+++ b/mmaction/datasets/rawframe_dataset.py
@@ -68,7 +68,7 @@ class RawframeDataset(BaseDataset):
         filename_tmpl (str): Template for each filename.
             Default: 'img_{:05}.jpg'.
         with_offset (bool): Determines whether the offset information is in
-            ann_file. If not, frame_idx starts from 1. Default: False.
+            ann_file. Default: False.
         multi_class (bool): Determines whether it is a multi-class
             recognition dataset. Default: False.
         num_classes (int): Number of classes in the dataset. Default: None.
--- a/mmaction/datasets/rawframe_dataset.py
+++ b/mmaction/datasets/rawframe_dataset.py
@@ -42,6 +42,21 @@ class RawframeDataset(BaseDataset):
         some/directory-5 295 3
         some/directory-6 121 3
 
+    Example of a with_offset annotation file (clips from long videos), each
+    line indicates the directory to frames of a video, the index of the start
+    frame, total frames of the video clip and the label of a video clip, which
+    are split with a whitespace.
+
+
+    .. code-block:: txt
+
+        some/directory-1 12 163 3
+        some/directory-2 213 122 4
+        some/directory-3 100 258 5
+        some/directory-4 98 234 2
+        some/directory-5 0 295 3
+        some/directory-6 50 121 3
+
 
     Args:
         ann_file (str): Path to the annotation file.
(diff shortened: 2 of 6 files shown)


tests/data/frame_test_list_with_offset.txt
```diff
@@ -0,0 +1,2 @@
+test_imgs 2 5 127
```
innerlee, line 1: are you defining a new dataset annotation format?

innerlee: what does each column mean?

kennymckormick: I add an optional format for ann_file, instead of 'frame_dir num_frame label[s]', it can be 'frame_dir **start_idx** num_frame label[s]'. My code is compatible with original format.

mmaction/datasets/pipelines/loading.py
```diff
@@ -591,7 +591,12 @@ def __call__(self, results):
         if results['frame_inds'].ndim != 1:
             results['frame_inds'] = np.squeeze(results['frame_inds'])
 
+        offset = 0
+        if 'offset' in results:
+            offset = results['offset']
```
dreamerlin, line 596:
```python
offset = results.get('offset', 0)
```

mmaction/datasets/rawframe_dataset.py
```diff
@@ -65,36 +67,49 @@ def __init__(self,
...
+                    item['total_frames'] = int(line_split[idx])
+                    idx += 1
+                label = [int(x) for x in line_split[idx:]]
+                assert len(label), 'missing label in line: {}'.format(line)
+                if len(label) == 1:
+                    item['label'] = label[0]
+                else:
+                    assert self.multi_class, (
+                        'found multiple labels in line'
+                        '{}, however multi_class == False'.format(line))
```
dreamerlin, line 106: use f-string format

mmaction/datasets/rawframe_dataset.py
```diff
@@ -65,36 +67,49 @@ def __init__(self,
...
-        self.filename_tmpl = filename_tmpl
 
     def load_annotations(self):
         """Load annotation file to get video information."""
         video_infos = []
         with open(self.ann_file, 'r') as fin:
             for line in fin:
                 line_split = line.strip().split()
-                if self.multi_class:
+                item = {}
```
dreamerlin, line 85: `item` -> `video_info`

mmaction/datasets/rawframe_dataset.py
```diff
@@ -65,36 +67,49 @@ def __init__(self,
...
+                idx += 1
+                if self.with_offset:
+                    item['offset'] = int(line_split[idx])
+                    item['total_frames'] = int(line_split[idx + 1])
+                    idx += 2
+                else:
+                    item['total_frames'] = int(line_split[idx])
+                    idx += 1
+                label = [int(x) for x in line_split[idx:]]
+                assert len(label), 'missing label in line: {}'.format(line)
```
dreamerlin, line 100: use f-string format.

mmaction/datasets/rawframe_dataset.py
```diff
@@ -52,6 +52,8 @@ class RawframeDataset(BaseDataset):
             Default: False.
         filename_tmpl (str): Template for each filename.
             Default: 'img_{:05}.jpg'.
+        with_offset (bool): Determines whether the offset information is in
+            ann_file. If not, frame_idx starts from 1. Default: False.
```
dreamerlin, line 56: Need to give example in docstring for untrimmed datasets annotation like #line22 - #line43

mmaction/datasets/rawframe_dataset.py
```diff
@@ -65,36 +67,49 @@ def __init__(self,
...
             for line in fin:
                 line_split = line.strip().split()
-                if self.multi_class:
+                item = {}
+                idx = 0
+                frame_dir = line_split[idx]
+                if self.data_prefix is not None:
+                    frame_dir = osp.join(self.data_prefix, frame_dir)
+                item['frame_dir'] = frame_dir
+                idx += 1
```
dreamerlin, line 91: idx in line 86 and line 91 can be directly hardcoded with comments.

mmaction/datasets/rawframe_dataset.py
```diff
@@ -65,36 +67,49 @@ def __init__(self,
...
+                if self.with_offset:
+                    item['offset'] = int(line_split[idx])
+                    item['total_frames'] = int(line_split[idx + 1])
+                    idx += 2
+                else:
+                    item['total_frames'] = int(line_split[idx])
+                    idx += 1
+                label = [int(x) for x in line_split[idx:]]
```
dreamerlin, line 99: replace idx in line 95 and line 98 with `label_start_index=3` and `label_start_index=2`

mmaction/datasets/rawframe_dataset.py
kennymckormick: code changed

mmaction/datasets/rawframe_dataset.py
kennymckormick: I think using a floating idx is a better choice. Since the ann_file may contain many optional parts in the future, using a floating idx improves the scalability of the codebase. (Or each time you need to add a new part, you need to change lots of code).

mmaction/datasets/rawframe_dataset.py
kennymckormick: plz refer to previous comments

mmaction/datasets/rawframe_dataset.py
dreamerlin: It makes sense bur it is still better to add a comment for each idx change. like `# index for frame_dir`, `# index for offset and total frames`, `# index for label`, etc.

tests/data/frame_test_list_with_offset.txt
innerlee: what `num_frame` in this context mean? the length of the video or the length of the segment?

kennymckormick: The length of the video clip (which is part of the entire untrimmed video).

innerlee: I think we need a doc page to describe all possible annotation formats we used. @dreamerlin

dreamerlin: We can prepare a `annotation_description.md` in `docs/` to describe all possible annotation formats.

innerlee:
I'm thinking on some short descriptor of the annotation file which resembles the channel order description.

Say we let
* `X`: video
* `Y`: label
* `S`: snippet start index (begin with 0, this number is included)
* `D`: duration of snippet
* `L`: length of video

Then this can be represented as ` XSDY`. Notice the starting space ` `, which represents the delimiter.
Other formats could be `,XLY`, meaning comma delimited, giving video name, video length and label.

kennymckormick:
> I'm thinking on some short descriptor of the annotation file which resembles the channel order description.
> 
> Say we let
> 
> * `X`: video
> * `Y`: label
> * `S`: snippet start index (begin with 0, this number is included)
> * `D`: duration of snippet
> * `L`: length of video
> 
> Then this can be represented as `" XSDY"`. Notice the starting space ` `, which represents the delimiter.
> Other formats could be `",XLY"`, meaning comma delimited, giving video name, video length and label.
> 
> For one based frame datasets, one can add an extra column `S=1`, with format `" VSDL"`.
> We can deprecate the many options like start_index, etc.
> 
> cc. @dreamerlin @hellock @kennymckormick
> 
> and @JoannaLXY , could this fit into localization tasks? Anything else needed to be added?

That's a good idea, where should we place the description?

innerlee:
## Proposal 2

A second thought, maybe we have an easier solution, that is to unify all recognition (single class) annotation to 
`video start_index length class`

This should cover all use cases, right?

kennymckormick:
> ## Proposal 2
> A second thought, maybe we have an easier solution, that is to unify all recognition (single class) annotation to
> `video start_index length class`
> 
> This should cover all use cases, right?

I think the current design is OK since: 
1. To unify all recognition (single class) annotation, you need to change a lot of codes, like data preprocessing, dataset, etc. 
2. The current code is both forward compatible and backward compatible, so we don't need to make extra efforts in unifying annotations. And think about that case: maybe we need to add many unforeseen features in the future. For example, to support HVU, we need to write a dataset that supports multiple attributes (with action class, scene class, sentiment class, etc.). At that time, the efforts made now may become worthless。

innerlee:
The current design is not ideal because it is stacking many switches. What's worse is that those switches are not independent.

A better design is needed. Any idea?

kennymckormick:
@hellock @dreamerlin Any good idea about annotation design for recognition? I think json which contains a list of dictionaries (each represents a sample) is a good choice. Each dictionary looks like `dict(frame_dir='path', num_frame=100, label=1)`, etc. . I think this design is good since it can be extended easily to your need.

hellock:
I suggest refactoring as early as possible. When there are more users, the backward compatibility issue will be a technical debt. We can define a primary annotation format described in a  json file, which will be more extensible.

kennymckormick:
> I suggest refactoring as early as possible. When there are more users, the backward compatibility issue will be a technical debt. We can define a primary annotation format described in a json file, which will be more extensible.

You mean we can support both txtlist and json, while using json as the primary format?

innerlee:
Let's merge it for now, and leave the unification of data annotation to future.

PR #53, opened 2020-07-25
Add TIN model

dreamerlin:


mmaction/models/backbones/cuda_shift/setup.py
```diff
@@ -1,9 +1,24 @@
...
+import subprocess
 from setuptools import find_packages, setup
 
 import torch
 from torch.utils.cpp_extension import CUDAExtension
 
+
+def _find_cuda_home():
+    # guess rule 3 of torch.utils.cpp_extension
+    nvcc = subprocess.check_output(['which', 'nvcc']).decode().rstrip('\r\n')
```
innerlee, line 11: this function can be moved to util or mmcv

mmaction/models/backbones/cuda_shift/setup.py
```diff
@@ -1,9 +1,24 @@
...
+    cuda_home = os.path.dirname(os.path.dirname(nvcc))
+    print(f'find cuda home:{cuda_home}')
+    return cuda_home
+
+
+# remember to overwrite PyTorch auto-detected cuda home which
```
innerlee, line 17: what's the reason?

mmaction/models/backbones/cuda_shift/setup.py
```diff
@@ -19,7 +34,7 @@
 else:
     raise ValueError('CUDA is not available')
 
-extra_compile_args = {'cxx': []}  # '-fopenmp', '-std=c99']
+extra_compile_args = dict(cxx=[])  # ['-fopenmp', '-std=c99']
```
innerlee, line 37: what does the commented flags mean? and why they are comments?

mmaction/models/backbones/cuda_shift/make.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+python setup.py build develop --user
```
innerlee, line 2: not a good place for shell scripts

setup.cfg
```diff
@@ -19,6 +19,6 @@ line_length = 79
 multi_line_output = 0
 known_standard_library = pkg_resources,setuptools
 known_first_party = mmaction
-known_third_party = cv2,joblib,matplotlib,mmcv,numpy,pandas,pytest,scipy,seaborn,torch,torchvision
+known_third_party = cudashift,cv2,joblib,matplotlib,mmcv,numpy,pandas,pytest,scipy,seaborn,torch,torchvision
```
innerlee, line 22: is cudashift a third party?

configs/recognition/tin/tin_r50_1x1x8_40e_sthv1_rgb.py
```diff
@@ -0,0 +1,131 @@
...
+    dict(type='ToTensor', keys=['imgs', 'label'])
+]
+val_pipeline = [
+    dict(
+        type='SampleFrames',
+        clip_len=1,
+        frame_interval=1,
+        num_clips=8,
+        test_mode=True),
+    dict(type='FrameSelector'),
```
innerlee, line 54: Unrelated to this pr, but the name `FrameSelector` feels a little strange. Maybe rename it to something like `SelectFrames`?

BTW what's the difference between select frames and sample frames?

mmaction/models/backbones/cuda_shift/setup.py
dreamerlin: Actually, this function is a simplified version of [_find_cuda_home](https://github.com/pytorch/pytorch/blob/master/torch/utils/cpp_extension.py#L27-L54) in PyTorch, which uses 3 Guess for `CUDA_HOME`. For the customized one, it directly sets directory path to `CUDA` which supports `nvcc` command as `CUDA_HOME`, which may be more reasonable.
The detailed description has been add as a comment.

mmaction/models/backbones/cuda_shift/setup.py
dreamerlin: Because PyTorch determines the `CUDA_HOME` using this priority: `os.environ['CUDA_HOME']` > path to `CUDA` supporting `nvcc` > `'/usr/local/cuda'`, although the first two guess should be the same in most cases. Here, it prefers using the second guess.

mmaction/models/backbones/cuda_shift/setup.py
dreamerlin: Unused compilation switch

setup.cfg
dreamerlin: I can not get rid of it. : )

configs/recognition/tin/tin_r50_1x1x8_40e_sthv1_rgb.py
dreamerlin: `sample frames` is to sample index from frames. `select frame` is to pick frame using index.

mmaction/models/backbones/cuda_shift/setup.py
innerlee: okay, not sure why this matters

configs/recognition/tin/tin_r50_1x1x8_40e_sthv1_rgb.py
innerlee: What is "pick"?

mmaction/models/backbones/cuda_shift/setup.py
innerlee: why change the cuda finding logic from pytorch? does it actually matters? or it was a temp workaround on the original authors machine?

setup.cfg
innerlee: it is an extension. extension does not need to be wrapped as a third-party lib, right? i guess mmdet has lots of extensions

mmaction/models/backbones/cuda_shift/setup.py
dreamerlin: I have confirmed it with the author, he said they are the same in most cases. 
But I think the CUDA_HOME of nvcc is the one we expected.
Maybe I can try it with the PyTorch `_find_cuda_home`

innerlee: yeah if pytorch's cuda home works on our cluster, then it's better to remove this custom finding logic. if there are flaws in pytorch's logic, we can patch upstream

configs/recognition/tin/tin_r50_1x1x8_40e_sthv1_rgb.py
dreamerlin: It means selecting specific frames using the fixed index after sampling.

innerlee: Okay. Select and Sample are too close in meaning. Maybe we can rename it to PickFrames. "Pick" somehow reminds of "cherry-pick", which means "choose and take only (the most beneficial or profitable items, opportunities, etc.) from what is available."

mmaction/models/backbones/resnet_tin.py
```diff
@@ -0,0 +1,370 @@
...
+    Then compute the weighted sum along with the temporal dimension.
+
+    Args:
+        data (torch.Tensor): Split data for certain group in shape
+            [N, num_segments, C, H, W].
+        offset (torch.Tensor): Data offsets for this group data in shape
+            [N, num_segments].
+    """
+    # [N, num_segments, C, H, W]
+    n, t, c, h, w = data.shape
```
innerlee, line 24: use upper case for n, t, c, h, w

we can use lowercase as the iteration variable

mmaction/models/backbones/resnet_tin.py
```diff
@@ -0,0 +1,370 @@
...
+
+        Args:
+            x (torch.Tensor): The input data.
+
+        Returns:
+            torch.Tensor: The output of the module.
+        """
+        # calculate weight
+        # [N, C, T]
+        n, _, t = x.shape
```
innerlee, line 131: same here, uppercase N and T

mmaction/models/backbones/resnet_tin.py
```diff
@@ -0,0 +1,370 @@
...
+
+        self.conv = nn.Conv1d(in_channels, groups, 3, padding=1)
+
+        self.init_weights()
+
+    def init_weights(self):
+        """Initiate the parameters either from existing checkpoint or from
+        scratch."""
+        # we set the initial bias of the convolution
+        # layer to 0, and the final initial output will be 1.0
```
innerlee, line 117: what is "final initial" output?

dreamerlin: image
This comment is the same with the original repo and the paper. It means the initialized conv layer will output 1.0

innerlee: what's the initial weights?

dreamerlin: The conv weights are initialized by default using pytorch `kaiming_uniform_` or loading from ckpt, the codes only changes the bias as the paper said.

innerlee (approved):

PR #57, opened 2020-07-26
add data benchmark for tsn

kennymckormick:
Detailed data benchmark for TSN with two data formats, two training augmentations, and two testing protocols (8 combinations).


configs/recognition/tsn/README.md
```diff
@@ -26,6 +26,21 @@
 
 Here, We use [1: 1] to indicate that we combine rgb and flow score with coefficients 1: 1 to get the two-stream prediction (without applying softmax).
 
+### Kinetics-400 Data Benchmark (8-gpus, ResNet50, ImageNet pretrain; 3 segments)
+
```
innerlee, line 30: pls add a short description, explaining what "data benchmark" means. and explain which variations are compared?

configs/recognition/tsn/README.md
```diff
@@ -26,6 +26,23 @@
...
+| [tsn_fix_resize_train](/configs/recognition/tsn/data_benchmark/tsn_fix_resize_train.py) |    340x256     |   RandomResizedCrop   |   25x10 frames   |  70.11   |  89.01   | [ckpt]() | [log]() | [json]() |
+|                              x                               |    340x256     |   RandomResizedCrop   |   25x3 frames    |  69.95   |  89.02   |    x     |    x    |    x     |
+| [tsn_se_multi_train](/configs/recognition/tsn/data_benchmark/tsn_se_multi_train.py) | short-side 320 |    MultiScaleCrop     |   25x10 frames   |  70.32   |  89.25   | [ckpt]() | [log]() | [json]() |
+|                              x                               | short-side 320 |    MultiScaleCrop     |   25x3 frames    |  70.54   |  89.39   |    x     |    x    |    x     |
+| [tsn_se_resize_train](/configs/recognition/tsn/data_benchmark/tsn_se_resize_train.py) | short-side 320 |   RandomResizedCrop   |   25x10 frames   |  70.44   |  89.23   | [ckpt]() | [log]() | [json]() |
+|                              x                               | short-side 320 |   RandomResizedCrop   |   25x3 frames    |  70.91   |  89.51   |    x     |    x    |    x     |
+
+Notes:
+
```
innerlee, line 43: Suggestion: 

* put this paragraph between the section title and the table, and remove the `Notes:`
* what does the `se` in the config name represents? very cryptic
* for the 320p data, how about in addition running a setting without the first resize augmentation (e.g. https://github.com/open-mmlab/mmaction2/pull/57/files#diff-addd50e00eeedda0c1f02c685ac94a1eR32), since scaling to height 256 defeats the purpose of being "short-side". You may use the lazy augmentation (as in https://github.com/open-mmlab/mmaction2/blob/master/configs/recognition/i3d/i3d_r50_fast_32x2x1_100e_kinetics400_rgb.py#L41). The lazy op would keep as much info as possible.

kennymckormick: Hi, innerlee. This [line] (https://github.com/open-mmlab/mmaction2/pull/57/files#diff-addd50e00eeedda0c1f02c685ac94a1eR32) actually indicates scaling shortedge to 256, so the meaning is correct. The lazy option is False by default, and I will run some experiments with lazy=True, to see if higher resolution makes any difference.

innerlee: Okay, so the meaning of scale is not (w, h), but some kind of scale bounds

innerlee: What does `se` mean

kennymckormick: In the latest commit, I add the line: ' In config names, **fix**(fix_ratio) and **se**(shortedge) denotes 340x256 data and short-edge 320px data respectively.' to explain the meaning of **fix** and **se**. Is it OK to use them in config names to keep them short and concise?

innerlee: long config names are okay. it's better to be able to guess the meaning of each part. so, shortedge is better than se, and fixratio is better than fix.

innerlee: btw in the past, we used 320p to denote short-side 320, eg https://github.com/open-mmlab/mmaction2/blob/master/configs/recognition/tsn/tsn_r50_320p_1x1x3_110e_kinetics400_flow.py

kennymckormick: Thats OK. I think maybe I can use 320p and 340x256 directly in the config name.

innerlee: that would be nice

innerlee:
Two more comments:
* what does the `_train` suffix mean? i guess our configs defaults to training configs
* could you provide the logs and ckpts? maybe @dreamerlin could help with the uploading

kennymckormick:
> Two more comments:
> 
> * what does the `_train` suffix mean? i guess our configs defaults to training configs
> * could you provide the logs and ckpts? maybe @dreamerlin could help with the uploading

1. The config names is further updated to align with the default format @dreamerlin.
2. The logs and ckpts have been shared to @dreamerlin, I will add links once ready.

PR #60, opened 2020-07-26
Support hmdb51 dataset preparation

dreamerlin:


tools/data/hmdb51/download_annotations.sh
```diff
@@ -0,0 +1,20 @@
+#! /usr/bin/bash env
```
innerlee, line 1: this line looks suspicious. isn't it `#!/usr/bin/env bash`?

tools/data/hmdb51/download_annotations.sh
```diff
@@ -0,0 +1,20 @@
...
+  echo "${DATA_DIR} does not exist. Creating";
+  mkdir -p ${DATA_DIR}
+fi
+
+cd ${DATA_DIR}
+wget http://serre-lab.clps.brown.edu/wp-content/uploads/2013/10/test_train_splits.rar --no-check-certificate
+
+# sudo apt-get install unrar
+unrar x test_train_splits.rar
+rm test_train_splits.rar
```
innerlee, line 15: it is very likely that user's pc does not have unrar installed. as a consequence, the newly downloaded data will be deleted without decompression. ref https://stackoverflow.com/questions/3474526/stop-on-first-error

tools/data/hmdb51/download_annotations.sh
dreamerlin: Change it to `#!/usr/bin/env bash`. Same changes in other scripts are made in https://github.com/open-mmlab/mmaction2/pull/62

tools/data/hmdb51/download_annotations.sh
dreamerlin: use `set -e` to make bash exit the script if any command fails

PR #69, opened 2020-07-29
Fix the bug of generating HMDB51 class index file.

yzfly:
1. list.sort()  function will return nothing, so the `class_list` will be none. Use sorted() instead.

2. The offical HMDB51 annotations file is [action]_test_split[1-3].txt  corresponding to three splits reported in the paper.

### DIFF
--- a/tools/data/parse_file_list.py
+++ b/tools/data/parse_file_list.py
@@ -373,7 +373,7 @@ def parse_hmdb51_split(level):
         frame_path = 'data/hmdb51/rawframes'
         annotation_dir = 'data/hmdb51/annotations'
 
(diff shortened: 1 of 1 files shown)


tools/data/parse_file_list.py
```diff
@@ -388,7 +388,7 @@ def generate_class_index_file():
             with open(train_file_template.format(i), 'w') as fout:
                 content = []
                 for class_name in class_dict:
-                    filename = class_name + f'_train_split{i}.txt'
+                    filename = class_name + f'_test_split{i}.txt'
```
innerlee, line 391: double check this line

PR #70, opened 2020-07-30
Polish data preparation codes

dreamerlin:
This PR polishes data preparation codes and has tested it to ensure its Correctness.

### DIFF
--- a/tools/data/parse_file_list.py
+++ b/tools/data/parse_file_list.py
@@ -7,7 +7,6 @@ import os.path as osp
 
 
 def parse_directory(path,
-                    key_func=lambda x: osp.basename(x),
                     rgb_prefix='img_',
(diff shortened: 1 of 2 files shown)


tools/data/parse_file_list.py
```diff
@@ -378,49 +382,41 @@ def generate_class_index_file():
         with open(class_index_file, 'w') as f:
             content = []
             for class_id, class_name in enumerate(class_list):
+                # like `ClassInd.txt` in UCF-101, the class_id begins with 1
```
innerlee, line 385: if class id begins with 1, then num class in the config file should be 52?

same question for ucf101

innerlee: okay never mind, the id subtracts 1 later

PR #71, opened 2020-07-31
correct the sthv1/sthv2 rawframes filelist generate command

yzfly:
tools/data/build_file_list.py only support two format ['rawframes', 'videos'], add missing 's' for the sthv1/sthv2 rawframes filelist generate command.
`parser.add_argument(
        '--format',
        type=str,
        default='rawframes',
        choices=['rawframes', 'videos'],
        help='data format')`


innerlee:
thanks!

PR #72, opened 2020-08-01
Support to output video or gif file in demo.py

dreamerlin:
* Support output video or gif in `demo.py`.
* Add output demo file (`demo.gif` and `demo_out.mp4`).


innerlee:
demo.gif is too large. pls reduce the size below 1.5M

demo/demo.py
```diff
@@ -18,10 +20,84 @@ def parse_args():
...
+    duration_video_clip = video_clips.duration
+    text_clips = TextClip(label, fontsize=font_size, color=font_color)
+    text_clips = (
+        text_clips.set_position(
+            ('right', 'bottom'),
+            relative=True).set_duration(duration_video_clip))
+
+    video_clips = CompositeVideoClip([video_clips, text_clips])
+
+    out_type = osp.splitext(osp.basename(out_filename))[1][1:]
```
innerlee, line 94: no need to call basename before splitext

demo/demo.py
```diff
@@ -18,10 +20,84 @@ def parse_args():
         help='whether to use rawframes as input')
     parser.add_argument(
         '--device', type=str, default='cuda:0', help='CPU/CUDA device option')
+    parser.add_argument(
+        '--fps', default=30, help='fps value of the output video')
```
innerlee, line 24: fps should be None. if the input is video, then probe the input fps.

dreamerlin: Modify the description to "specify fps value of the output video when using rawframes to generate file"

innerlee:
mmaction2_overview.gif is much larger than the current demo.gif

dreamerlin:
> mmaction2_overview.gif is much larger than the current demo.gif

They are both about 1.5MB.

PR #80, opened 2020-08-03
Add `target_resolution` in demo and fix the fps params

dreamerlin:
* Add `target_resolution` args which supports to resize the frames in a video when using it as a input. It can help reduce the size of the output file or generate high resolution video.
* Fix bugs for type in args.

### DIFF
--- a/demo/demo.py
+++ b/demo/demo.py
@@ -34,6 +34,13 @@ def parse_args():
         default='white',
         help='font color of the label test in output video')
     parser.add_argument(
+        '--target-resolution',
+        nargs=2,
+        default=None,
+        type=int,
+        help='Set to (desired_height, desired_width) for resizing the frames '
+        'when using a video as input')
+    parser.add_argument(
         '--resize-algorithm',
         default='bicubic',
         help='resize algorithm applied to generate video')
@@ -48,6 +55,7 @@ def get_output(video_path,
                fps=30,
                font_size=20,
                font_color='white',
+               target_resolution=None,
                resize_algorithm='bicubic',
                use_frames=False):
     """Get demo output using ``moviepy``.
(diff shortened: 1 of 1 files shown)


demo/demo.py
```diff
@@ -65,9 +75,14 @@ def get_output(video_path,
         fps (int): Number of picture frames to read per second. Default: 30.
         font_size (int): Font size of the label. Default: 20.
         font_color (str): Font color of the label. Default: 'white'.
-        resize_algorithm (str): The algorithm used for resizing.
-            Default: 'bicubic'. For more information,
-            see https://ffmpeg.org/ffmpeg-scaler.html.
+        target_resolution (None | tuple[int, int]): Same as that in
+            ``VideoFileClip``. Set to (desired_height, desired_width) to have
```
innerlee, line 79: resolutions are often in the order of width, height. e.g. ffmpeg uses that order.

dreamerlin: refer to [this](https://github.com/Zulko/moviepy/blob/master/moviepy/video/io/VideoFileClip.py#L38-L42), and it revert the resolution [here](https://github.com/Zulko/moviepy/blob/2cd208c3791931aaea21b0ae278469bdfa57a96d/moviepy/video/io/ffmpeg_reader.py#L40-L41)

innerlee: we do not need to expose the order of `VideoFileClip` here. moviepy is an implementation detail, that can be replaced at any time

innerlee: considering moviepy has installation issues both on windows and unix, there is a great probability that we replace it with opencv

dreamerlin: It will be an alternative for future plan.

dreamerlin:
Test successfully by the examples ofthis [docs](https://github.com/open-mmlab/mmaction2/pull/81)

demo/demo.py
```diff
@@ -65,9 +75,14 @@ def get_output(video_path,
         fps (int): Number of picture frames to read per second. Default: 30.
         font_size (int): Font size of the label. Default: 20.
         font_color (str): Font color of the label. Default: 'white'.
-        resize_algorithm (str): The algorithm used for resizing.
-            Default: 'bicubic'. For more information,
-            see https://ffmpeg.org/ffmpeg-scaler.html.
+        target_resolution (None | tuple[int, int]): Same as that in
+            ``VideoFileClip``. Set to (desired_width desired_height) to have
```
innerlee, line 79: Set to (desired_width desired_height) to have resized frames. If either dimension is None, the frames are resized by keeping the existing aspect ratio. Default: None.

demo/demo.py
```diff
@@ -65,9 +75,14 @@ def get_output(video_path,
...
+            ffmpeg resize the frames before returning them. If either
+            dimension is None, the frames are resized by keeping the existing
+            aspect ratio. Default: None.
+        resize_algorithm (str): Same as that in ``VideoFileClip``.
```
innerlee, line 83: support "bicubic", "bilinear", "neighbor", "lanczos", etc. For more information, see https://ffmpeg.org/ffmpeg-scaler.html

PR #86, opened 2020-08-06
Yield TSN Feature for BSN / BMN

kennymckormick:
reopen last pr for yielding TSN feature


mmaction/datasets/pipelines/loading.py
```diff
@@ -178,6 +178,61 @@ def __call__(self, results):
...
 
+@PIPELINES.register_module()
+class UntrimSampleFrames(object):
+    """Sample frames from the untrimmed video.
+
+    Required keys are "filename", added or modified keys are "total_frames",
+    "frame_inds", "frame_interval" and "num_clips".
+
+    Args:
+        clip_len (int): The length of sampled clips. Default: 1. (5 for Flow)
```
innerlee, line 189: what does `(5 for Flow)` mean? does the default changes to 5 when the input is flow?

mmaction/datasets/pipelines/loading.py
```diff
@@ -178,6 +178,61 @@ def __call__(self, results):
         return results
 
 
+@PIPELINES.register_module()
+class UntrimSampleFrames(object):
```
innerlee, line 182: Untrim is verb so this name causes ambiguity. How about something like `SampleUntrimmedFrames` or `SampleFramesFromUntrimmed` or `UntrimmedSampleFrames`

mmaction/datasets/pipelines/loading.py
```diff
@@ -178,6 +178,61 @@ def __call__(self, results):
...
+    def __call__(self, results):
+        """Perform the SampleFrames loading.
+
+        Args:
+            results (dict): The resulting dict to be modified and passed
+                to the next transform in pipeline.
+        """
+        if 'total_frames' not in results:
+            # TODO: find a better way to get the total frames number for video
+            video_reader = mmcv.VideoReader(results['filename'])
```
innerlee, line 213: can you guarantee that the input is video?

mmaction/datasets/pipelines/loading.py
innerlee: A question, can we reuse the class SampleFrames and remove this class?

tools/tsn_feature_extract.py
```diff
@@ -0,0 +1,140 @@
...
+    parser.add_argument('--clip-len', type=int, default=1, help='clip length')
+    parser.add_argument('--modality', default='RGB')
+    parser.add_argument('--ckpt', help='checkpoint for feature extraction')
+    parser.add_argument(
+        '--part',
+        type=int,
+        default=0,
+        help='which part of dataset to forward(alldata[part::tot])')
+    parser.add_argument(
+        '--tot', type=int, default=1, help='how many parts exist')
```
innerlee, line 35: tot?

tools/anet_feature_prepare.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Returns:
+        np.ndarray: The pooled feature with shape num_proposals x feature_dim.
+    """
+    if len(data) == 1:
+        return np.concatenate([data] * num_proposals)
+    x_range = list(range(len(data)))
+    f = scipy.interpolate.interp1d(x_range, data, axis=0)
+    eps = 1e-4
+    st, ed = eps, len(data) - 1 - eps
```
dreamerlin, line 43: `st` -> `start`, `ed` -> `end`

tools/tsn_feature_extract.py
kennymckormick: means total, changed to total now

mmaction/datasets/pipelines/loading.py
kennymckormick: No, it means when we use UntrimSampleFrames to sample flow from untrimmed videos, we need to set it 5. Since we have wrote it in the config file, I decided to remove it to reduce ambiguity.

mmaction/datasets/pipelines/loading.py
kennymckormick: UntrimmedSampleFrames will be a good name.

kennymckormick: I think we would better not to reuse SampleFrames since SampleFrames is about sampling **num_clips** clips  from the video. In our case, the **num_clips** is unknown for an untrimmed video. We need to sample unknown numbers of clips from an untrimmed video of unknown length with a given distance.

mmaction/datasets/pipelines/loading.py
kennymckormick: Yes, it's clear that only when you use rawframe_dataset, the key 'total_frames' will appear in the video_info. Otherwise, the input is a video.

dreamerlin: It seems not right, the class `XXX_Init` will add key `total_frames` in it. you can rebase master to get the latest codes.

dreamerlin:
there are lint problems now. you may need to manually check all files by 'pre-commit run --all-files'

mmaction/datasets/pipelines/loading.py
kennymckormick: I see. Now that the 'total_frames' key becomes essential, removed the check.

PR #89, opened 2020-08-07
Move `start_index` from `SampleFrames` to dataset level

dreamerlin:
This PR moves `start_index` from `SampleFrames` to dataset level, which also has been tested in training.


mmaction/datasets/pipelines/loading.py
```diff
@@ -17,17 +17,13 @@ class SampleFrames(object):
     """Sample frames from the video.
 
     Required keys are "filename", added or modified keys are "total_frames",
-    "frame_inds", "frame_interval" and "num_clips".
+    "frame_inds", "frame_interval" "start_index" and "num_clips".
```
innerlee, line 20: does start_index belong to required keys?

mmaction/datasets/pipelines/loading.py
```diff
@@ -53,7 +48,6 @@ def __init__(self,
         self.clip_len = clip_len
         self.frame_interval = frame_interval
         self.num_clips = num_clips
-        self.start_index = start_index
```
innerlee, line 56: for easier migration for users, we may check the use of start_index here and raise a detailed error message, describing how to modify code (or link to a faq section of docs)

dreamerlin: It is hard to check `start_index` here, since we can not get the result dict here to judge whether it's a video or rawframe. Besides, For videos the `start_index` should be 0, but for rawframes it depends on the files themselves to assign 0 or 1 to `start_idx`.

innerlee: Check it here, in the __init__

because we no longer support start_index in init, any usage of it should be warned

innerlee:
en?

dreamerlin:
> en?

I don't know what happened. I just delete the branch on my fork. : )

PR #90, opened 2020-08-07
use `xxInit()` method to get total_frames

dreamerlin:
This pr uses `xxInit()` method to get total_frames, it is a better way to get the total frames number for video.


innerlee:
change docstring

PR #91, opened 2020-08-07
Use `RawFrameDecode` to replace `FrameSelector`

dreamerlin:
This pr uses `SelectFrames` to replace `FrameSelector` and make `FrameSelector` deprecated.


innerlee:
PickFrames?

innerlee:
Still not quite understand the functionality of PickFrames.

* What does it do? Load and decode image frames?
* Does it needed for video loaders?

Here is my guess:

For videos, we need XXXInit to get number of frames. So the pipeline looks like `XXXInit -> SampleFrames -> XXXDecode`
For RawFrames, number of frames is loaded from annotations, so we do not need YYYInit. Then the logic goes like `SampleFrames -> LoadAndDecodeFrames`

So, is PickFrames some kind of `LoadAndDecodeFrames`?

dreamerlin:
> RawFrameDecode

Agree

innerlee:
Cool. Now I'm thinking that we can add a NEWS.md file in the root folder, in which we record all breaking changes. A good example to follow: https://github.com/JuliaLang/julia/blob/master/NEWS.md

PR #93, opened 2020-08-08
bug fix: add missing args in `load_checkpoint()`

dreamerlin:


mmaction/models/backbones/resnet.py
```diff
@@ -508,7 +508,8 @@ def init_weights(self):
                     self.pretrained, strict=False, logger=logger)
             else:
                 # ours
-                load_checkpoint(self.pretrained, strict=False, logger=logger)
+                load_checkpoint(
+                    self, self.pretrained, strict=False, logger=logger)
```
hellock, line 512: Do not let users specify the argument `torchvision_pretrain`. It can be inferred from the checkpoint.

innerlee: how to infer

PR #99, opened 2020-08-10
Add BSN / BMN results with feature extracted by mmaction

kennymckormick:
Add BSN / BMN results with feature extracted by mmaction


dreamerlin:
LGTM

kennymckormick:
@innerlee seems OK

PR #100, opened 2020-08-10
Fix quote issues and rename fast as dense

SuX97:
Fix quote issues and rename fast as dense


innerlee:
this commit has already live in master. closing

PR #109, opened 2020-08-12
support multi class in tsm head

sjtuytc:


dreamerlin:
Thanks for your PR, but you still need to click [here](https://cla-assistant.io/open-mmlab/mmaction2?pullRequest=109) to sign CLA before we can accept your contribution.

mmaction/models/heads/tsm_head.py
```diff
@@ -34,8 +34,9 @@ def __init__(self,
                  dropout_ratio=0.8,
                  init_std=0.001,
                  is_shift=True,
+                 multi_class=False,
```
dreamerlin, line 37: Please add description for this args in the docstring like [this](https://github.com/open-mmlab/mmaction2/blob/master/mmaction/models/heads/base.py#L40).

sjtuytc:
Signed just now.

PR #110, opened 2020-08-12
Fix bug in work_dir.

SuX97:
The `args.work_dir` will never be None, it will only be `''` an empty str. When training without specific work_dir, it results in lots of logs and ckpts saved in $MMACTION2_HOME.

### DIFF
--- a/tools/slurm_train.sh
+++ b/tools/slurm_train.sh
@@ -5,12 +5,11 @@ set -x
 PARTITION=$1
 JOB_NAME=$2
 CONFIG=$3
-WORK_DIR=$4
 GPUS=${GPUS:-8}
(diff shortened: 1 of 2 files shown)


innerlee:
image

SuX97:
But in `slurm_train.sh`, it will be 
image. @innerlee

innerlee:
that is a bug of slurm_train.sh

innerlee:
2

SuX97:
I vote for change the slurm scripts, just remove the work_dir to PY_ARGS as the user wishes. @dreamerlin

dreamerlin:
Well, we can try plan 2 to see what will happen.

innerlee:
Another choice: abandon `slurm_train.sh` and adopt `slurm_train`

PR #111, opened 2020-08-12
Update sth doc

tony2016uestc:
Remove 'filename_tmpl' in sthv2 configs and add the step of encoding sthv1 videos.


tools/data/sthv1/preparing_sthv1.md
```diff
@@ -19,11 +19,11 @@ cat 20bn-something-something-v1-?? | tar zx
 cd $MMACTION2/tools/data/sthv1/
 ```
 
-For users who only want to use RGB frames, you can skip to step 4 to generate file lists in the format of rawframes. Since the prefix of official JPGs is "%05d.jpg" (e.g., "00001.jpg"), you have to add "filename_tmpl='{:05}.jpg'" to the dict of data.train, data.val and data.test in the config files related with sthv1 like this:
+For users who only want to use RGB frames, you can skip to step 5 to generate file lists in the format of rawframes. Since the prefix of official JPGs is "%05d.jpg" (e.g., "00001.jpg"), we add "filename_tmpl='{:05}.jpg'" to the dict of data.train, data.val and data.test in the config files related with sthv1 like this:
```
innerlee, line 22: quote the ```  `data.train`  ``` and ``` `data.val` ``` etc.

PR #112, opened 2020-08-12
support multi class in tsm head

sjtuytc:
support multi class in TSM head


dreamerlin:
Thanks for your PR. BTW, it is ok to directly push your commit in the existing pr and no need to delete it, since the changes will be updated automatically. : )

PR #114, opened 2020-08-13
Update SampleProposalFrames.

tony2016uestc:
Replace `out_props` with `out_proposals`.


innerlee:
may detect the usage of `out_props` and raise an error about how to migrate. it will be good to link to this pr in the error message

dreamerlin:
Why this pr is a breaking?

innerlee:
the changed key may break user's code

PR #126, opened 2020-08-14
Update docs

dreamerlin:
This PR updates docs according to [this change](https://github.com/open-mmlab/mmaction2/pull/110/files)


innerlee:
if .`/slurm_train` was used, all those issue would not have happened

PR #127, opened 2020-08-15
Bug fix in CSN mode setting

dreamerlin:
This PR fixes bug, which affects the testing, in `def train(self, mode=True):` in CSN. The modification has been well tested.


innerlee:
may add a test case that capture the prior failure

PR #130, opened 2020-08-16
Use `fuse_conv_bn` from mmcv

dreamerlin:


innerlee:
any approximate numbers on the speedup?

PR #135, opened 2020-08-17
Multigrid: Add long-cycle.

SuX97:
Add long-cycle.

### DIFF
--- a/mmaction/core/multi_grid_utils.py
+++ b/mmaction/core/multi_grid_utils.py
@@ -1,0 +1,252 @@
+import numpy as np
+import torch.nn as nn
+from mmcv.runner import HOOKS, Hook, LrUpdaterHook
+from mmcv.runner.hooks.lr_updater import StepLrUpdaterHook
+from torch.nn.modules.utils import _ntuple
+
+from mmaction.datasets.builder import build_dataloader
+from mmaction.datasets.pipelines.augmentations import Resize
+from mmaction.datasets.pipelines.loading import SampleFrames
+from mmaction.models.common import SubBatchBN3d
+from ..utils import get_root_logger
+
+
+def modify_num_splits(module, num_splits):
+    """Recursively modify the number of splits of subbn3ds in module.
+
+    Args:
+        module (nn.Module): The module to be modified.
+        num_splits (int): The targeted number of splits.
+
+    Returns:
+        int: The number of subbn3d modules modified.
+    """
+    count = 0
+    for child in module.children():
+        if isinstance(child, SubBatchBN3d):
+            new_split_bn = nn.BatchNorm3d(child.num_features *
+                                          num_splits).cuda()
+            child.num_splits = num_splits
+            child.split_bn = new_split_bn
+            count += 1
+        else:
+            count += modify_num_splits(child, num_splits)
+    return count
+
+
+@HOOKS.register_module()
+class RelativeStepLrUpdaterHook(LrUpdaterHook):
+    """RelativeStepLrUpdaterHook.
+
+    Args:
+        ori_step (list[int]): Same as that of mmcv.
+        new_step (list[int]): The real steps caused by multi-grid
+            training.
+        step_lr_ratio (list[int]): The extra factor to be multiplied
+            at each new step.
+        gamma (float): Same as that of mmcv.
+        **kwargs (dict): Same as that of mmcv.
+    """
+
+    def __init__(self, runner, step, lrs, **kwargs):
+        super().__init__(**kwargs)
+        assert len(step) == (len(lrs) - 1)
+        self.step = step
+        self.lrs = lrs
+        super().before_run(runner)
+
+    def get_lr(self, runner, base_lr):
+        progress = runner.epoch if self.by_epoch else runner.iter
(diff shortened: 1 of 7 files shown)


mmaction/core/multi_grid_utils.py
```diff
@@ -0,0 +1,252 @@
+import numpy as np
+import torch.nn as nn
+from mmcv.runner import HOOKS, Hook, LrUpdaterHook
+from mmcv.runner.hooks.lr_updater import StepLrUpdaterHook
+from torch.nn.modules.utils import _ntuple
+
+from mmaction.datasets.builder import build_dataloader
+from mmaction.datasets.pipelines.augmentations import Resize
+from mmaction.datasets.pipelines.loading import SampleFrames
+from mmaction.models.common import SubBatchBN3d
```
dreamerlin, line 10:
```
from mmaction.datasets import build_dataloader, Resize, SampleFrames
from mmaction.models import SubBatchBN3d
```

mmaction/core/multi_grid_utils.py
```diff
@@ -0,0 +1,252 @@
...
+    """RelativeStepLrUpdaterHook.
+
+    Args:
+        ori_step (list[int]): Same as that of mmcv.
+        new_step (list[int]): The real steps caused by multi-grid
+            training.
+        step_lr_ratio (list[int]): The extra factor to be multiplied
+            at each new step.
+        gamma (float): Same as that of mmcv.
+        **kwargs (dict): Same as that of mmcv.
```
dreamerlin, line 48: Args doesn't correspond with the __init__

mmaction/core/multi_grid_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        for i in range(len(self.step)):
+            if progress < self.step[i]:
+                return self.lrs[i]
+
+
+class MultiGridHook(Hook):
+    """MultiGridHook.
+
+    https://arxiv.org/abs/1912.00998.
+    """
```
dreamerlin, line 69:
```
A multigrid method for efficiently training video models

This hook defines multigrid training schedule and update cfg accordingly, which is proposed in 
`A Multigrid Method for Efficiently Training Video Models <https://arxiv.org/abs/1912.00998>`_
```

mmaction/core/multi_grid_utils.py
```diff
@@ -0,0 +1,252 @@
...
+            total_epochs += epochs
+            final_schedule.append((s[0], s[1], int(round(total_epochs))))
+        self._print_schedule(final_schedule)
+        return final_schedule
+
+    def _print_schedule(self, schedule):
+        """logging the schedule."""
+        self.logger.info('\tLongCycleId\tBase shape\tEpochs\t')
+        for s in schedule:
+            self.logger.info('\t{}\t{}\t{}\t'.format(s[0], s[1], s[2]))
```
dreamerlin, line 216: f-string

mmaction/core/multi_grid_utils.py
```diff
@@ -0,0 +1,252 @@
...
+        gamma (float): Same as that of mmcv.
+        **kwargs (dict): Same as that of mmcv.
+    """
+
+    def __init__(self, runner, step, lrs, **kwargs):
+        super().__init__(**kwargs)
+        assert len(step) == (len(lrs) - 1)
+        self.step = step
+        self.lrs = lrs
+        super().before_run(runner)
```
dreamerlin, line 56: Is this line necessary? [It](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/epoch_based_runner.py#L103) seems the `before_run` will be automatically call.

SuX97: Yeah, it is because necessary attributes such as `self.base_lr` is initiated in `before_run` of a LrHook, but when I swap one with the new one. The new one will not be initialized.

mmaction/core/multi_grid_utils.py
SuX97: I have updated it in the big PR.

mmaction/core/multi_grid_utils.py
SuX97: Resize, SampleFrames can only be imported from mmactoin.datasets.pipeliens. But yeah, will revise. BTW is that a convtion to not import from the last layer?

mmaction/core/multi_grid_utils.py
SuX97: Another solution is just move the contents in `before_run` to `__init__`, it will be better to understand.

mmaction/core/multi_grid_utils.py
dreamerlin: Commonly import from outer layer which is no more than 2 level/

mmaction/core/multi_grid_utils.py
dreamerlin: Oh I know, it may be casued by the order of the hook. Since the original `StepLrHook` called `before_run` first, and the new one `LrHook` will not call `before_run`, resulting in the no initialization. So, just `super().before_run(runner)` is OK.

PR #138, opened 2020-08-19
Fix random unittest failure for BaseDataset

dreamerlin:
This PR resolves a random failure for unittest caused by out of bound in `BaseDataset`.

### DIFF
--- a/tests/test_data/test_dataset.py
+++ b/tests/test_data/test_dataset.py
@@ -287,7 +287,6 @@ class TestDataset(object):
             data_prefix=self.data_prefix,
             start_index=3)
         assert len(video_dataset) == 2
-        assert type(video_dataset[0]) == dict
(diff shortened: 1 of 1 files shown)


innerlee:
why it fails? any related ci logs?

dreamerlin:
> why it fails? any related ci logs?

[CI logs](https://github.com/open-mmlab/mmaction2/runs/991748813).  It fails because the valid `start_index` for our video data for unittest is 0. In `def test_base_dataset()`, it just uses `start_index=3` to test assignment instead of pipeline.

innerlee:
okay so test throw for index 0 and test the type for index 3?

dreamerlin:
> okay so test throw for index 0 and test the type for index 3?

It fails due to out of bound.

innerlee (approved):

PR #150, opened 2020-08-28
Move identical `evaluate` into BaseDataset

SuX97:
These two are pretty much the same, and will not differ even when the user writes a new dataset. When the user writes a new evaluation metric, he/she only needs to modify it in BaseDataset.

### DIFF
--- a/mmaction/datasets/base.py
+++ b/mmaction/datasets/base.py
@@ -4,8 +4,10 @@ from abc import ABCMeta, abstractmethod
 
 import mmcv
 import torch
+from mmcv.utils import print_log
 from torch.utils.data import Dataset
 
+from ..core import mean_average_precision, mean_class_accuracy, top_k_accuracy
 from .pipelines import Compose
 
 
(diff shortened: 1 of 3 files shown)


mmaction/datasets/base.py
```diff
@@ -90,19 +92,83 @@ def load_json_annotations(self):
                 video_infos[i]['label'] = video_infos[i]['label'][0]
         return video_infos
 
-    @abstractmethod
-    def evaluate(self, results, metrics, logger):
-        """Evaluation for the dataset.
+    def evaluate(self,
+                 results,
+                 metrics='top_k_accuracy',
+                 topk=(1, 5),
```
innerlee, line 98: putting topk at base dataset is not very good

SuX97: agree. topk is a option for 'top_k_accuracy'. I think a better way is to warp metrics as a class, and use config dict to initialize it like
```python
metrics = build_metrics(dict(type='top_k_accuracy'), topk=(1, 5))
```

innerlee:
Fix https://github.com/open-mmlab/mmpose/issues/90

PR #151, opened 2020-08-28
Fix the multi-class support in a cleaner way.

SuX97:
Move multi-class and stuff to **kwargs, do not need to add it once a new head appears.

### DIFF
--- a/mmaction/models/heads/slowfast_head.py
+++ b/mmaction/models/heads/slowfast_head.py
@@ -18,6 +18,8 @@ class SlowFastHead(BaseHead):
         spatial_type (str): Pooling type in spatial dimension. Default: 'avg'.
         dropout_ratio (float): Probability of dropout layer. Default: 0.8.
         init_std (float): Std value for Initiation. Default: 0.01.
+        kwargs (dict, optional): Any keyword argument to be used to initialize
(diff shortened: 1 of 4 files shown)


mmaction/models/heads/slowfast_head.py
```diff
@@ -26,9 +28,10 @@ def __init__(self,
                  loss_cls=dict(type='CrossEntropyLoss'),
                  spatial_type='avg',
                  dropout_ratio=0.8,
-                 init_std=0.01):
+                 init_std=0.01,
+                 **kwargs):
 
-        super().__init__(num_classes, in_channels, loss_cls)
+        super().__init__(num_classes, in_channels, loss_cls, kwargs)
```
innerlee, line 34: **

innerlee:
has the multi-class option been tested?

SuX97:
> has the multi-class option been tested?

[Here](https://github.com/open-mmlab/mmaction2/blob/master/tests/test_models/test_head.py#L119). Only tested on TSNHead, but I think it will work as long as all heads are written in the same way.

innerlee:
fair

innerlee (approved):

PR #152, opened 2020-08-29
Flow extraction doc

kennymckormick:
@dreamerlin  This PR to generalize `tools/flow_extraction.py` for flow extraction of one or several videos, docs also added.

### DIFF
--- a/docs/data_preparation.md
+++ b/docs/data_preparation.md
@@ -76,6 +76,25 @@ The recommended practice is
 ln -s ${YOUR_FOLDER} $MMACTION2/data/$DATASET/rawframes
 ```
 
+#### Alternative to denseflow
+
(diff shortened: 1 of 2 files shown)


docs/data_preparation.md
```diff
@@ -76,6 +76,25 @@ The recommended practice is
 ln -s ${YOUR_FOLDER} $MMACTION2/data/$DATASET/rawframes
 ```
 
+#### Alternative to denseflow
+
+In case your device doesn't fulfill the installation requirement of [denseflow](https://github.com/open-mmlab/denseflow)(like Nvidia driver version), or you just want to see some quick demos about flow extraction, we provide a python script `tools/flow_extraction.py` as an alternative to denseflow. You can use it for rgb framesand optical flow extraction from one or several videos. Note that the speed of the script is much slower than denseflow, since it runs optical flow algorithms on CPU.
```
innerlee, line 81: frames and

innerlee (approved):

PR #167, opened 2020-09-08
Change default value of `norm_eval` to False

kennymckormick:
Actually, the default value of norm_eval should be False. 
Thus if a user forgets to set this arg, the training is still OK.


dreamerlin (approved):

innerlee:
does this change affect any existing config file?

kennymckormick:
> does this change affect any existing config file?

Seems not, all configs have set norm_eval explicitly

PR #175, opened 2020-09-13
Add bninception.

tony2016uestc:


dreamerlin:
Unittest is needed.

dreamerlin:
@hellock Should the `BNInception` model be reconstruct by `ConvModule` ？

dreamerlin:
@tony2016uestc The `BNInception` should be also tested in `test_backbone` to check the input and output shape.
.

PR #177, opened 2020-09-14
Use interpolate instead of slice to support onnx.

SuX97:
image

Slice and interpolate are equivalent.

### DIFF
--- a/mmaction/models/backbones/resnet3d_slowfast.py
+++ b/mmaction/models/backbones/resnet3d_slowfast.py
@@ -458,14 +458,14 @@ class ResNet3dSlowFast(nn.Module):
             samples extracted by the backbone.
         """
         x_slow = nn.functional.interpolate(
-            x, scale_ratio=(1.0 / self.resample_rate, 1.0, 1.0))
+            x, scale_factor=(1.0 / self.resample_rate, 1.0, 1.0))
         x_slow = self.slow_path.conv1(x_slow)
         x_slow = self.slow_path.maxpool(x_slow)
 
         x_fast = nn.functional.interpolate(
             x,
-            scale_ratio=(1.0 / (self.resample_rate // self.speed_ratio), 1.0,
-                         1.0))
+            scale_factor=(1.0 / (self.resample_rate // self.speed_ratio), 1.0,
+                          1.0))
         x_fast = self.fast_path.conv1(x_fast)
         x_fast = self.fast_path.maxpool(x_fast)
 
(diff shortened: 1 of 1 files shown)


innerlee:
any speed difference?

SuX97:
> any speed difference?

Will test and report it soon.

SuX97:
> any speed difference?

slice using 0.13951969146728516
interpolate using 0.1295177936553955

Even a bit faster!

```
import time
import torch

a = torch.randn(8, 3, 32, 112, 112).cuda()
tik = time.time()
for trial in range(1000):
    result = a[:, :, ::4, :, :]
tok = time.time()
print(f'slice using {tok - tik}')

tik = time.time()
for trial in range(1000):
    result = torch.nn.functional.interpolate(a, scale_factor=(1.0 / 4.0, 1.0, 1.0))
tok = time.time()
print(f'interpolate using {tok - tik}')
```

mmaction/models/backbones/resnet3d_slowfast.py
```diff
@@ -457,11 +457,15 @@ def forward(self, x):
...
-        x_slow = x[:, :, ::self.resample_rate, :, :]
+        x_slow = nn.functional.interpolate(
+            x, scale_factor=(1.0 / self.resample_rate, 1.0, 1.0))
         x_slow = self.slow_path.conv1(x_slow)
         x_slow = self.slow_path.maxpool(x_slow)
 
-        x_fast = x[:, :, ::self.resample_rate // self.speed_ratio, :, :]
+        x_fast = nn.functional.interpolate(
+            x,
+            scale_factor=(1.0 / (self.resample_rate // self.speed_ratio), 1.0,
```
innerlee, line 467: is `//` necessary?

SuX97: By default of the paper and PySlowfast, it is unnecessary, as in all cases it can divide (32 / 4, 32 / 8, 64 /4, 64 / 8).

innerlee: should `self.resample_rate >= self.speed_ratio` be asserted?

SuX97: Yes, resample_rate should always > speed_ratio, otherwise it will slice with step 0. But I think it will occur frequently, only when user 魔改 model, and the error log: ValueError: slice step cannot be zero is easy to track.

innerlee:
okay, this is fine for the default interpolation mode 'nearest'.

innerlee (approved):

innerlee:
maybe be explicit about the interpolation mode is better

PR #191, opened 2020-09-17
[Enhance] Async inference of webcam demo

dreamerlin:
This PR makes `webcam_demo.py` not only supports `TSN`, but also other recognizers which need enough frames for inference at the beginning to do real-time recognition. Also, this PR updates the docs for how to make efficient real-time action recognition.

### DIFF
--- a/demo/webcam_demo.py
+++ b/demo/webcam_demo.py
@@ -3,6 +3,7 @@ from collections import deque
 from operator import itemgetter
 
 import cv2
+import mmcv
 import numpy as np
 import torch
 from mmcv.parallel import collate, scatter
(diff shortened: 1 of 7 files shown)


demo/webcam_demo.py
```diff
@@ -51,26 +58,38 @@ def parse_args():
...
     while True:
         ret, frame = camera.read()
         # BGR to RGB
         windows.append(np.array(frame[:, :, ::-1]))
         if data['img_shape'] is None:
             data['img_shape'] = frame.shape[:2]
 
+        if need_preparation and not is_prepared:
+            prog_bar.update()
+            if len(windows) != sample_length:
```
innerlee, line 79: how about fill the whole buffer with the first frame, so that no preparation is needed?

dreamerlin: Done.

demo/webcam_demo.py
```diff
@@ -21,6 +21,11 @@
     'PyAVDecode', 'RawFrameDecode', 'FrameSelector'
 ]
 
+NEED_STUFF = [
+    'ResNetTSM', 'ResNetTIN', 'ResNet3dCSN', 'ResNet3d', 'ResNet2Plus1d',
```
innerlee, line 25: can this "stuff" thing be avoided completely?

dreamerlin: It can be avoided if we "stuff" for all models including `TSN`.

innerlee: so why not

innerlee (approved):

PR #204, opened 2020-09-23
[Improve] Remove redundant `dist_utils.py`

dreamerlin:
Remove redundant `dist_utils.py`, which has been used in [mmcv](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/fp16_utils.py#L200-L241)


innerlee:
deleting is always good

innerlee (approved):

PR #211, opened 2020-09-26
[Docs] Add Customize runtime tutorial

dreamerlin:


docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,321 @@
+# Tutorial 5: Customize Runtime Settings
+
+## Customize Optimization Settings
+
+### Customize optimizer supported by PyTorch
+
+We already support to use all the optimizers implemented by PyTorch, and the only modification is to change the `optimizer` field of config files.
+For example, if you want to use `ADAM` (note that the performance could drop a lot), the modification could be as the following.
```
innerlee, line 8: remove this note

innerlee: why use the all CAPITAL `ADAM` here and use `Adam` in the config?

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,321 @@
+# Tutorial 5: Customize Runtime Settings
```
innerlee, line 1: could you explain the `Runtime` a little bit?

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,321 @@
...
+
+We already support to use all the optimizers implemented by PyTorch, and the only modification is to change the `optimizer` field of config files.
+For example, if you want to use `ADAM` (note that the performance could drop a lot), the modification could be as the following.
+
+```python
+optimizer = dict(type='Adam', lr=0.0003, weight_decay=0.0001)
+```
+
+To modify the learning rate of the model, the users only need to modify the `lr` in the config of optimizer.
+The users can directly set arguments following the [API doc](https://pytorch.org/docs/stable/optim.html?highlight=optim#module-torch.optim) of PyTorch.
```
innerlee, line 15: this line is pretty vague. could u provide an example?

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,321 @@
...
+@OPTIMIZERS.register_module()
+class MyOptimizer(Optimizer):
+
+    def __init__(self, a, b, c):
+
+```
+
+#### 2. Add the optimizer to registry
+
+To find the above module defined above, this module should be imported into the main namespace at first. There are two options to achieve it.
```
innerlee, line 41: two ways to achieve it

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,321 @@
...
+- Use `custom_imports` in the config to manually import it
+
+```python
+custom_imports = dict(imports=['mmaction.core.optimizer.my_optimizer'], allow_failed_imports=False)
+```
+
+The module `mmaction.core.optimizer.my_optimizer` will be imported at the beginning of the program and the class `MyOptimizer` is then automatically registered.
+Note that only the package containing the class `MyOptimizer` should be imported. `mmaction.core.optimizer.my_optimizer.MyOptimizer` **cannot** be imported directly.
+
+Actually users can use a totally different file directory structure using this importing method, as long as the module root can be located in `PYTHONPATH`.
```
innerlee, line 61: Improve the writing of this line

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+so that 1 epoch for training and 1 epoch for validation will be run iteratively.
+
+**Note**:
+
+1. The parameters of model will not be updated during val epoch.
+2. Keyword `total_epochs` in the config only controls the number of training epochs and will not affect the validation workflow.
+3. Workflows `[('train', 1), ('val', 1)]` and `[('train', 1)]` will not change the behavior of `EvalHook` because `EvalHook` is called by `after_train_epoch` and validation workflow only affect hooks that are called through `after_val_epoch`.
+Therefore, the only difference between `[('train', 1), ('val', 1)]` and ``[('train', 1)]` is that the runner will calculate losses on validation set after each training epoch.
```
innerlee, line 183: unpaired ``` ` ```

docs/tutorials/customize_runtime.md
dreamerlin: Just delete it.

docs/tutorials/customize_runtime.md
dreamerlin: An Example is given as the following

docs/tutorials/customize_runtime.md
dreamerlin: A description is given as the following.

docs/tutorials/customize_runtime.md
innerlee: cool

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,328 @@
...
+        policy='cyclic',
+        target_ratio=(0.85 / 0.95, 1),
+        cyclic_times=1,
+        step_ratio_up=0.4,
+    )
+    ```
+
+## Customize Training Schedules
+
+By default we use step learning rate with 1x schedule, this calls [`StepLRHook`](https://github.com/open-mmlab/mmcv/blob/f48241a65aebfe07db122e9db320c31b685dc674/mmcv/runner/hooks/lr_updater.py#L153) in MMCV.
```
innerlee, line 147: we do not use the term `1x schedule`, right?

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,328 @@
...
+## Customize Workflow
+
+Workflow is a list of (phase, epochs) to specify the running order and epochs. By default it is set to be
+
+```python
+workflow = [('train', 1)]
+```
+
+which means running 1 epoch for training.
+Sometimes user may want to check some metrics (e.g. loss, accuracy) about the model on the validate set.
```
innerlee, line 176: this line invites trouble.
it implies that when you want to do evaluation, then the de facto way is to use the val workflow.
on the contrary, the val workflow is discouraged, and we only implement the eval hooks.

Need to revise this section to reflect the reality

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,328 @@
...
+
+```python
+mmcv_hooks = [
+    dict(type='MMCVHook', a=a_value, b=b_value, priority='NORMAL')
+]
+```
+
+### Modify default runtime hooks
+
+There are some common hooks that are not registered through `custom_hooks`, they are
```
innerlee, line 283: then how are they registered?

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,328 @@
...
+Therefore, the only difference between `[('train', 1), ('val', 1)]` and ``[('train', 1)]`` is that the runner will calculate losses on validation set after each training epoch.
+
+## Customize Hooks
+
+### Customize self-implemented hooks
+
+#### 1. Implement a new hook
+
+There are some occasions when the users might need to implement a new hook. MMAction2 supports customized hooks in training (#xxx) since xxxx, thus the users could implement a hook directly in mmaction2
+or their mmaction2-based codebases and use the hook by only modifying the config in training. Before vxxxx, the users need to modify the code to get the hook registered before training starts.
```
dreamerlin, line 199: other

docs/tutorials/customize_runtime.md
```diff
@@ -0,0 +1,328 @@
...
+3. Workflows `[('train', 1), ('val', 1)]` and `[('train', 1)]` will not change the behavior of `EvalHook` because `EvalHook` is called by `after_train_epoch` and validation workflow only affect hooks that are called through `after_val_epoch`.
+Therefore, the only difference between `[('train', 1), ('val', 1)]` and ``[('train', 1)]`` is that the runner will calculate losses on validation set after each training epoch.
+
+## Customize Hooks
+
+### Customize self-implemented hooks
+
+#### 1. Implement a new hook
+
+There are some occasions when the users might need to implement a new hook. MMAction2 supports customized hooks in training (#xxx) since xxxx, thus the users could implement a hook directly in mmaction2
```
dreamerlin, line 198: MMAction2

docs/tutorials/customize_runtime.md
dreamerlin: Recommend users to use `EvalHook` at first and use `val` workflow as an alternative.

docs/tutorials/customize_runtime.md
dreamerlin: There are some common hooks that are not registered through `custom_hooks` but has been registered by default when importing MMCV

docs/tutorials/customize_runtime.md
innerlee, line 201: del L198-L199

docs/tutorials/customize_runtime.md
innerlee: yeah better

docs/tutorials/customize_runtime.md
innerlee: that would be good

innerlee (approved):

PR #212, opened 2020-09-26
[Enhance] Add `cfg-options` in argument

dreamerlin:


tools/bsn_proposal_generation.py
```diff
@@ -169,6 +169,10 @@ def main():
     print('Begin Proposal Generation Module')
     args = parse_args()
     cfg = mmcv.Config.fromfile(args.config)
+    # import modules from string list.
+    if cfg.get('custom_imports', None):
```
innerlee, line 173: repeating these three lines are so verbose. the mechanism should be improved.

### Proposal

add a new function `mmcv.Config.resolve_custom_imports`, which imports modules into `mmcv.Config.custom_imports`.
add an option in `mmcv.Config.from_file` to enable `resolve_custom_imports`.

After this, things will be:

```python
cfg = mmcv.Config.fromfile(args.config, resolve_custom_imports=True) # could be shorter if True is default
from mmcv.Config.custom_imports import *
```

Benefit: reduce 3 lines to 1 line

cc. @hellock

innerlee: or incorporate `import_modules_from_strings` into fromfile, and reduce 3 lines to 0 lines.

tools/test.py
```diff
@@ -41,6 +42,12 @@ def parse_args():
         help='tmp directory used for collecting results from multiple '
         'workers, available when gpu-collect is not specified')
     parser.add_argument('--options', nargs='+', help='custom options')
+    parser.add_argument(
+        '--cfg-options',
+        nargs='+',
+        action=DictAction,
+        help='override some settings in the used config, the key-value pair '
+        'in xxx=yyy format will be merged into config file.')
```
innerlee, line 50: use an example of overwriting two options would be great

tools/test.py
```diff
@@ -72,7 +79,10 @@ def merge_configs(cfg1, cfg2):
 def main():
     args = parse_args()
 
-    cfg = mmcv.Config.fromfile(args.config)
+    cfg = Config.fromfile(args.config)
+
+    if args.cfg_options is not None:
+        cfg.merge_from_dict(args.cfg_options)
```
innerlee, line 85: if `--cfg-options` defaults to an empty dict, can the above two lines be reduced into one line?

tools/test.py
dreamerlin: The example is provide in [merge_from_dict](https://github.com/open-mmlab/mmcv/blob/master/mmcv/utils/config.py#L406-L418)

dreamerlin: Example is added

tools/test.py
dreamerlin: set `default={}` now

innerlee (approved):

PR #213, opened 2020-09-26
[Enhance] Add config context to meta dict

dreamerlin:


tools/train.py
```diff
@@ -141,6 +141,8 @@ def main():
             mmaction_version=__version__ + get_git_hash(digits=7),
             config=cfg.text)
 
+    meta['config'] = cfg.pretty_text
```
innerlee, line 144: maybe it is more polite to sanitize paths in config.

since other repos clearly haven't concern privacy as much, let's merge it and fix it until someone actually raise an issue.

innerlee: what do u think, @dreamerlin , if u want to remove the folder structure, at least those contain user name

PR #222, opened 2020-09-28
Hvu datatools

kennymckormick:


tools/data/hvu/generate_file_list.py
```diff
@@ -0,0 +1,151 @@
+import argparse
```
dreamerlin, line 1: Can this part be moved to (parse_file_list.py)[https://github.com/open-mmlab/mmaction2/blob/master/tools/data/parse_file_list.py] ?

kennymckormick: It would be hard to do so: 1. The format of the datalist of HVU dataset is specific and can only be handled with this piece of code. Besides, the code can not be used on any other dataset. 2. Since the directory `tools` is not under the mmaction2 repository, I can not reuse the code in `parse_file_list.py` (You see that I copied the function `parse_directory` into this file). Thus I think it would be better to keep this code in the `hvu` directory.

dreamerlin: All right. But can we use `from ..parse_file_list import parse_directory` to avoid rewriting the `parse_directory`

tools/data/hvu/generate_file_list.py
```diff
@@ -0,0 +1,151 @@
...
+    data_list = open(args.input_csv).readlines()
+    data_list = [x.strip().split(',') for x in data_list[1:]]
+
+    if args.mode == 'videos':
+        downloaded = os.listdir(args.src_dir)
+        downloaded = [x.split('.')[0] for x in downloaded]
+        downloaded_set = set(downloaded)
+    else:
+        parse_result = parse_directory(args.src_dir)
+        downloaded_set = set(parse_result.keys())
```
dreamerlin, line 126: `downloaded_set = set(parse_result)`

tools/data/hvu/download_annotations.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
```
dreamerlin, line 2: Better to use `set -e` at the beginning. ref https://stackoverflow.com/questions/3474526/stop-on-first-error

tools/data/hvu/generate_file_list.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def parse_line(line):
+        tags, youtube_id, start, end = line
+        start, end = int(float(start) * 10), int(float(end) * 10)
+        newname = '{}_{:06d}_{:06d}'.format(youtube_id, start, end)
```
dreamerlin, line 131: Use f-string

tools/data/hvu/preparing_hvu.md
```diff
@@ -0,0 +1,84 @@
...
+
+```shell
+bash generate_videos_filelist.sh
+# execute the command below when rawframes are ready
+bash generate_rawframes_filelist.sh
+```
+
+## Step 5. Folder Structure
+
+After the whole data pipeline for Kinetics-400 preparation.
```
dreamerlin, line 56: HVU

tools/data/hvu/preparing_hvu.md
```diff
@@ -0,0 +1,84 @@
...
+you can get the rawframes (RGB + Flow), videos and annotation files for Kinetics-400.
```
dreamerlin, line 57: HVU

tools/data/hvu/preparing_hvu.md
```diff
@@ -0,0 +1,84 @@
...
+
+In the context of the whole project (for Kinetics-400 only), the *minimal* folder structure will look like:
+(*minimal* means that some data are not necessary: for example, you may want to evaluate kinetics-400 using the original video format.)
```
dreamerlin, line 60: HVU

dreamerlin:
please also add a link [here](https://github.com/open-mmlab/mmaction2/blob/master/docs/data_preparation.md)

tools/data/hvu/download.py
```diff
@@ -0,0 +1,199 @@
+# This scripts is copied from
+# https://github.com/activitynet/ActivityNet/blob/master/Crawler/Kinetics/download.py  # noqa: E501
```
innerlee, line 2: add license info. style example: https://github.com/open-mmlab/mmpose/blob/master/mmpose/core/post_processing/post_transforms.py#L2-L3

tools/data/hvu/download.py
```diff
@@ -0,0 +1,199 @@
...
+import argparse
+import glob
+import os
+import shutil
+import subprocess
+
+import mmcv
+
+import ssl  # isort:skip
```
innerlee, line 11: is `# isort:skip` necessary?

innerlee:
A side note: we have copied the download script from activitynet two times. Can them be merged to one?

tools/data/hvu/generate_file_list.py
kennymckormick: When executing the import `from ..parse_file_list import parse_directory`, python will throw the ValueError: attempted relative import beyond top-level package .

kennymckormick:
> A side note: we have copied the download script from activitynet two times. Can they be merged to one?

Seems that python scripts in tools can not reuse codes else where. Besides that, the 3 download.py do share some common part, but not the majority (less than 50% I guess).

tools/data/hvu/generate_file_list.py
dreamerlin: Maybe we can try `sys.path.append` ?

kennymckormick: Is there any walkaround to avoid adding 3rdparty to setup.cfg? If so, it will be OK. (I have tried isort:skip and it didn't work). Using sys.path.append is OK. Howevere, the problem is: if u try to `import  parse_file_list`, 3rdparty will be added.

tools/data/hvu/parse_tag_list.py
```diff
@@ -0,0 +1,15 @@
+import mmcv
+
+tag_list = '../../../data/hvu/annotations/hvu_categories.csv'
+
+lines = open(tag_list).readlines()
```
dreamerlin, line 5:
```
with open(tag_list, "r") as f:
    lines = f.readlines()
```

tools/data/hvu/generate_file_list.py
dreamerlin: All right

dreamerlin:
Need to resolve the conflict and update the changelog.

tools/data/hvu/parse_tag_list.py
kennymckormick: Why performing this change? The change makes codes longer. btw, since we only read the tag_list (not write it), we do not need to close it. I think the advantage of using `with` is to automatically handle file closing when writing to a file.

dreamerlin: Agree.

PR #232, opened 2020-10-02
[Improve] Change default value of `average-clips`

kennymckormick:
Set 'average_clips' in the test_cfg of each config file. 
You can still set this parameter during testing and it will override the default value in the config file.

### DIFF
--- a/configs/recognition/csn/ircsn_ig65m_pretrained_bnfrozen_r152_32x2x1_58e_kinetics400_rgb.py
+++ b/configs/recognition/csn/ircsn_ig65m_pretrained_bnfrozen_r152_32x2x1_58e_kinetics400_rgb.py
@@ -21,7 +21,7 @@ model = dict(
         init_std=0.01))
 # model training and testing settings
 train_cfg = None
-test_cfg = dict(average_clips=None)
+test_cfg = dict(average_clips='score')
 # dataset settings
 dataset_type = 'RawframeDataset'
 data_root = 'data/kinetics400/rawframes_train'
--- a/configs/recognition/csn/ircsn_ig65m_pretrained_bnfrozen_r152_32x2x1_58e_kinetics400_rgb.py
+++ b/configs/recognition/csn/ircsn_ig65m_pretrained_bnfrozen_r152_32x2x1_58e_kinetics400_rgb.py
@@ -21,7 +21,7 @@ model = dict(
         init_std=0.01))
 # model training and testing settings
 train_cfg = None
-test_cfg = dict(average_clips='score')
+test_cfg = dict(average_clips='prob')
 # dataset settings
 dataset_type = 'RawframeDataset'
 data_root = 'data/kinetics400/rawframes_train'
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,6 @@
 ## Changelog
 
-### v0.8.0
+### v0.8.0 (master)
 
 **Improvements**
 - Set default values of 'average_clips' in each config file so that there is no need to set it explicitly during testing in most cases ([#232](https://github.com/open-mmlab/mmaction2/pull/232))
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.8.0
+
+**Improvements**
+- Set default values of 'average_clips' in each config file so that there is no need to set it explicitly during testing in most cases ([#232](https://github.com/open-mmlab/mmaction2/pull/232))
+
+
 ### v0.7.0 (30/9/2020)
 
 **Highlights**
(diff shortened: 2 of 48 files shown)


dreamerlin:
It should be 'prob' rather than 'score' for 3D model. 'prob' will get higher performance for 3D.

configs/recognition/csn/ircsn_ig65m_pretrained_bnfrozen_r152_32x2x1_58e_kinetics400_rgb.py
```diff
@@ -21,7 +21,7 @@
         init_std=0.01))
 # model training and testing settings
 train_cfg = None
-test_cfg = dict(average_clips=None)
+test_cfg = dict(average_clips='score')
```
innerlee, line 24: 'score' means logit?

innerlee:
Do u need to change model zoo performance when changing configs from None to 'something'?

configs/recognition/csn/ircsn_ig65m_pretrained_bnfrozen_r152_32x2x1_58e_kinetics400_rgb.py
kennymckormick: > The default value of `average_clips` should be "prob" rather than "score" for 3D model. "prob" will get higher performance for 3D. And for 2D recognizers, just set `None` for `average_clips` is OK.

Sorry that I made the mistake, the default value should be 'prob'. I will update this PR.

kennymckormick: Yes.

kennymckormick:
> The default value of `average_clips` should be "prob" rather than "score" for 3D model. "prob" will get higher performance for 3D. And for 2D recognizers, just set `None` for `average_clips` is OK.

Sorry that I made the mistake, the default value should be 'prob'.

kennymckormick:
> Do u need to change model zoo performance when changing configs from None to 'something'?

Emm, Yes. Since the original default value of 'average_clips' is 'score', so some ckpts trained by me should be re-tested. @SuX97, did you explicitly set 'average_clips' to 'prob' during testing previously?

dreamerlin:
> > Do u need to change model zoo performance when changing configs from None to 'something'?
> 
> Emm, Yes. Since the original default value of 'average_clips' is 'score', so some ckpts trained by me should be re-tested. @SuX97, did you explicitly set 'average_clips' to 'prob' during testing previously?

The performance of the ckpt is test by setting 'average_clips' to 'prob', so there is no need to re-test.

kennymckormick:
> > > Do u need to change model zoo performance when changing configs from None to 'something'?
> > 
> > 
> > Emm, Yes. Since the original default value of 'average_clips' is 'score', so some ckpts trained by me should be re-tested. @SuX97, did you explicitly set 'average_clips' to 'prob' during testing previously?
> 
> The performance of the ckpt is test by setting 'average_clips' to 'prob', so there is no need to re-test.

My bad, 3D checkpoints trained by me is tested with default setting (`average_clips = 'score'`). So the reported performance might be a little bit lower. I will create another PR to fix them.

innerlee:
the perf table can be updated here too, so that configs and reported accuracies are always matched.

docs/changelog.md
```diff
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.8.0
```
innerlee, line 3: `### v0.8.0 (master)`

innerlee:
As long as the settings are consistent with the reported accuracies in modelzoo, i'm ok.

I'm a little confused by the use of None. Here are two possible meanings
* invalid. this argument does not make sense for this method
* don't care and use whatever the default value.

If it belongs to the second case, then the direction of this pr is correct: configs should be clear and concrete.

dreamerlin:
> As long as the settings are consistent with the reported accuracies in modelzoo, i'm ok.
> 
> I'm a little confused by the use of None. Here are two possible meanings
> 
> * invalid. this argument does not make sense for this method
> * don't care and use whatever the default value.
> 
> If it belongs to the second case, then the direction of this pr is correct: configs should be clear and concrete.

It belongs to the second case. For 3D recognizers, the `average_clips` matters since there is no `consensus` op to average the segments and `average_clip` does this thing. For some 2D recognizers, if it uses `consensus` op and doesn't reshape for temporal dimension (like TSM in `3crop + twice sample` setting, which will be supported soon), the `average_clips` doesn't matter.

kennymckormick:
> > As long as the settings are consistent with the reported accuracies in modelzoo, i'm ok.
> > I'm a little confused by the use of None. Here are two possible meanings
> > 
> > * invalid. this argument does not make sense for this method
> > * don't care and use whatever the default value.
> > 
> > If it belongs to the second case, then the direction of this pr is correct: configs should be clear and concrete.
> 
> It belongs to the second case. For 3D recognizers, the `average_clips` matters since there is no `consensus` op to average the segments and `average_clip` does this thing. For some 2D recognizers, if it uses `consensus` op and doesn't reshape for temporal dimension (like TSM in `3crop + twice sample` setting, which will be supported soon), the `average_clips` doesn't matter.

I agree. @innerlee

innerlee (approved):

PR #234, opened 2020-10-04
Make version.py independent

innerlee:


mmaction/version.py
```diff
@@ -1,7 +1,18 @@
 # Copyright (c) Open-MMLab. All rights reserved.
-from mmcv import parse_version_info
 
 __version__ = '0.7.0'
-short_version = __version__
+
+
+def parse_version_info(version_str):
```
innerlee, line 6:
```python
In [11]: def parse_version_info(version_str):
    ...:     return tuple(int(x) if x.isdigit() else x for x in versi
    ...: on_str.replace('rc','.rc').split('.'))
    ...: 
    ...: 

In [12]: parse_version_info('0.2.16') == (0, 2, 16)
Out[12]: True

In [13]: parse_version_info('1.2.3') == (1, 2, 3)
Out[13]: True

In [14]: parse_version_info('1.2.3rc0') == (1, 2, 3, 'rc0')
Out[14]: True

In [15]: parse_version_info('1.2.3rc1') == (1, 2, 3, 'rc1')
Out[15]: True

In [16]: parse_version_info('1.0rc0') == (1, 0, 'rc0')
Out[16]: True
```

tests/test_version_utils.py
```diff
@@ -0,0 +1,9 @@
+from mmaction.version import parse_version_info
+
+
+def test_parse_version_info():
```
hellock, line 4: This unit test has been done in mmcv.

hellock (approved):

PR #238, opened 2020-10-05
[Feature] Add AVA preparation

dreamerlin:


tools/data/build_rawframes.py
```diff
@@ -152,6 +152,11 @@ def parse_args():
         default='avi',
         choices=['avi', 'mp4', 'webm'],
         help='video file extensions')
+    parser.add_argument(
+        '--mixed-ext',
+        action='store_true',
+        default=False,
```
kennymckormick, line 158: No need to add default here since u use `store_true`.

tools/data/ava/preparing_ava.md
```diff
@@ -0,0 +1,131 @@
...
+First of all, you can run the following script to prepare annotations.
+
+```shell
+bash download_annotations.sh
+```
+
+This command will download `ava_v2.1.zip` for AVA `v2.1` annotation. If you need the AVA `v2.2` annotation, you can try the following script.
+
+```shell
+VERSIO=2.2 bash download_annotations.sh
```
kennymckormick, line 17: VERSION

tools/data/ava/preparing_ava.md
```diff
@@ -0,0 +1,131 @@
...
+
+Then, use the following script to prepare videos. The codes are adapted from the [official crawler](https://github.com/cvdfoundation/ava-dataset).
+Note that this might take a long time.
+
+```shell
+bash download_videos.sh
+```
+
+Note that if you happen to have sudoer or have [GNU parallel](https://www.gnu.org/software/parallel/) on your machine,
+you can speed up the procedure by downloading in parallel.
```
kennymckormick, line 30: What about wrapping parallel downloading in a python script? I have checked that `parallel` is not a default package on Linux. Seems other data tools in the repo use python to implement parallel downloading.

dreamerlin: Agree.

dreamerlin:
cc. @kennymckormick

tools/data/ava/preparing_ava.md
dreamerlin: Done

tools/data/ava/preparing_ava.md
```diff
@@ -0,0 +1,136 @@
...
+If you didn't install denseflow, you can still extract RGB frames using OpenCV by the following script, but it will keep the original size of the images.
+
+```shell
+bash extract_rgb_frames_opencv.sh
+```
+
+Or using ffmpeg to extract RGB frames by the following script.
+
+```shell
+bash extract_rgb_frames_ffmpeg.sh
```
kennymckormick, line 79: Since we provide 2 approaches for frame extraction, ideally, they should extract frames in the same fashion. However, I noticed that when using ffmpeg, the framerate was specified to be 30; when using opencv, the framerate was not specified. The 2 approaches are the same only if all videos in AVA are 30fps, I'm not sure if this is the case.

dreamerlin: `cut_videos.sh` will use `ffmpeg` to make the video 30 FPS. (`-r 30`)

kennymckormick: That would be fine.

kennymckormick (approved):

PR #241, opened 2020-10-09
[Improve] Twice Sample + 3 crops for TSM

dreamerlin:
Add another testing setting for TSM and other models


kennymckormick:
After reading the `SampleFrames` function, I found that the `twice_sample` option can be very similar to doubling the num_clips. For example, For example, if u use num_clips=2 + twice_sample, to sample 40 frames, the index you get is [0, 10, 20, 30]; if u use num_clips=4, the index you get is [5, 15, 25, 35]. Seems there is no big difference.

What about testing the performance of twice_sample and doubling num_clips. If the performances are similar, we can remove this option.

dreamerlin:
> After reading the `SampleFrames` function, I found that the `twice_sample` option can be very similar to doubling the num_clips. For example, For example, if u use num_clips=2 + twice_sample, to sample 40 frames, the index you get is [0, 10, 20, 30]; if u use num_clips=4, the index you get is [5, 15, 25, 35]. Seems there is no big difference.
> 
> What about testing the performance of twice_sample and doubling num_clips. If the performances are similar, we can remove this option.

The `twice_sample` does matter when there is a `reshape` after the ClassHead score as that in TSM. I have conduct a comparison experiment and found that `num_clip = 16` will drop the performance a lot (45+ -> 35+) and `num_clip=8 + twice sample` will improve the performance when using the same ckpt and reshape. 

The reason for this difference may consist in the sample strategy. when we training a TSM with `num_segs = 8`, it mean we use 8 segments to represent the whole video. so for the test, we may need to try the same sample strategy to represent the video. 

For straightly `num_clips=16`, suppose the frame index `[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16]` at the beginning, it will be converted to `[1,2,3,4,5,6,7,8], [9,10,11,12,13,14,15,16]`. Obviously, both of the batches can not represent the whole video for 8 segments.
But for `num_clips=8 + twice_sample`,  the frame index will be `[1,3,5,7,9,11,13,15,2,4,6,8,10,12,14,16]` at the beginning and be converted to `[1,3,5,7,9,11,13,15], [2,4,6,8,10,12,14,16]`. both of the batches can not represent the whole video for 8 segments.

kennymckormick:
> > After reading the `SampleFrames` function, I found that the `twice_sample` option can be very similar to doubling the num_clips. For example, For example, if u use num_clips=2 + twice_sample, to sample 40 frames, the index you get is [0, 10, 20, 30]; if u use num_clips=4, the index you get is [5, 15, 25, 35]. Seems there is no big difference.
> > What about testing the performance of twice_sample and doubling num_clips. If the performances are similar, we can remove this option.
> 
> The `twice_sample` does matter when there is a `reshape` after the ClassHead score as that in TSM. I have conduct a comparison experiment and found that `num_clip = 16` will drop the performance a lot (45+ -> 35+) and `num_clip=8 + twice sample` will improve the performance when using the same ckpt and reshape.
> 
> The reason for this difference may consist in the sample strategy. when we training a TSM with `num_segs = 8`, it mean we use 8 segments to represent the whole video. so for the test, we may need to try the same sample strategy to represent the video.
> 
> For straightly `num_clips=16`, suppose the frame index `[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16]` at the beginning, it will be converted to `[1,2,3,4,5,6,7,8], [9,10,11,12,13,14,15,16]` after reshape. Obviously, both of the batches can not represent the whole video for 8 segments.
> But for `num_clips=8 + twice_sample`, the frame index will be `[1,3,5,7,9,11,13,15,2,4,6,8,10,12,14,16]` at the beginning and be converted to `[1,3,5,7,9,11,13,15], [2,4,6,8,10,12,14,16]` after reshape. both of the batches can reasonably represent the whole video for 8 segments.
> 
> This impact is particularly huge when we use `twice_sample + 3 crops` in `sth` dataset, since there are many actions like `move in and out`, `push in and out`, etc.

U r right. I didn't consider the characteristic of TSM and made the mistake.

dreamerlin:
Ping @innerlee

innerlee:
later

innerlee (approved):

PR #250, opened 2020-10-14
[Improvement] set default batch_size of evaluation and testing to 1

dreamerlin:
This PR sets default batch_size of evaluation and testing as 1 to avoid OOM and manual config changes in most cases, since it used the training `batch_size` as default when `videos_per_gpu=cfg.data.get('videos_per_gpu', 2)`

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -53,10 +53,6 @@ def parse_args():
         'in xxx=yyy format will be merged into config file. For example, '
         "'--cfg-options model.backbone.depth=18 model.backbone.with_cp=True'")
     parser.add_argument(
-        '--multi-batches',
-        action='store_true',
-        help='whether to perform multi batches inference')
-    parser.add_argument(
         '--average-clips',
         choices=['score', 'prob', None],
         default=None,
@@ -132,8 +128,7 @@ def main():
     # build the dataloader
     dataset = build_dataset(cfg.data.test, dict(test_mode=True))
     dataloader_setting = dict(
-        videos_per_gpu=cfg.data.get('videos_per_gpu', 1)
-        if args.multi_batches else 1,
+        videos_per_gpu=cfg.data.get('videos_per_gpu', 1),
         workers_per_gpu=cfg.data.get('workers_per_gpu', 0),
         dist=distributed,
         shuffle=False)
--- a/tools/test.py
+++ b/tools/test.py
@@ -128,7 +128,7 @@ def main():
     # build the dataloader
     dataset = build_dataset(cfg.data.test, dict(test_mode=True))
     dataloader_setting = dict(
-        videos_per_gpu=cfg.data.get('videos_per_gpu', 2),
+        videos_per_gpu=1,
         workers_per_gpu=cfg.data.get('workers_per_gpu', 0),
         dist=distributed,
         shuffle=False)
--- a/tools/test.py
+++ b/tools/test.py
@@ -129,7 +129,7 @@ def main():
     dataset = build_dataset(cfg.data.test, dict(test_mode=True))
     dataloader_setting = dict(
         videos_per_gpu=cfg.data.get('videos_per_gpu', 1),
-        workers_per_gpu=cfg.data.get('workers_per_gpu', 0),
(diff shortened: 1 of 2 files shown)


innerlee:
what does `most cases` refers to?

dreamerlin:
> what does `most cases` refers to?

It means the `cfg.data.videos_per_gpu` is set for training but too large for testing or evaluation (since the testing and evaluation may need more crops), which will cause OOM, and users may need to manually change the config setting for `cfg.data.videos_per_gpu` or `cfg.data.val_dataloader`. So we set the `videos_per_gpu` to 1 to avoid this case though it is slow.

innerlee:
but videos_per_gpu support val and test

innerlee:
see https://github.com/open-mmlab/mmaction2/blob/master/configs/localization/bmn/bmn_400x100_2x8_9e_activitynet_feature.py#L69

dreamerlin:
> see https://github.com/open-mmlab/mmaction2/blob/master/configs/localization/bmn/bmn_400x100_2x8_9e_activitynet_feature.py#L69

U mean it is better to change the config file by setting the `val_dataloader` or `test_dataloader` rather than use 1 batch_size as default?

innerlee:
>  but too large for testing or evaluation

sure, this solves the problem, isn't it?

tools/test.py
```diff
@@ -115,7 +115,7 @@ def main():
     # build the dataloader
     dataset = build_dataset(cfg.data.test, dict(test_mode=True))
     dataloader_setting = dict(
-        videos_per_gpu=cfg.data.get('videos_per_gpu', 2),
+        videos_per_gpu=1,
```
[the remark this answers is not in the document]
dreamerlin: https://github.com/open-mmlab/mmaction2/blob/master/tools/test.py#L47-L54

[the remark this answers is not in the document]
dreamerlin: `videos_per_gpu=cfg.data.get('videos_per_gpu', 1) if args.multi_batches else 1` ?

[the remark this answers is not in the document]
kennymckormick: That would be a good choice.

dreamerlin:
@kennymckormick @innerlee

tools/test.py
```diff
@@ -52,6 +52,10 @@ def parse_args():
         help='override some settings in the used config, the key-value pair '
         'in xxx=yyy format will be merged into config file. For example, '
         "'--cfg-options model.backbone.depth=18 model.backbone.with_cp=True'")
+    parser.add_argument(
+        '--multi-batches',
```
innerlee, line 56: what does multi-batch mean

dreamerlin: It will determine whether to set `batch` as `cfg.data.videos_per_gpu` first.

dreamerlin: But `cfg.data.videos_per_gpu` may raise OOM, so this args provide a option for users.

tools/test.py
```diff
@@ -128,7 +132,8 @@ def main():
     # build the dataloader
     dataset = build_dataset(cfg.data.test, dict(test_mode=True))
     dataloader_setting = dict(
-        videos_per_gpu=cfg.data.get('videos_per_gpu', 2),
+        videos_per_gpu=cfg.data.get('videos_per_gpu', 1)
+        if args.multi_batches else 1,
         workers_per_gpu=cfg.data.get('workers_per_gpu', 0),
```
kennymckormick, line 137: Better to set the default value of 'workers_per_gpu' as 1 (for better testing speed).

kennymckormick:
@innerlee This PR is of **high priority**, old codes have bugs when performing localization testing.

tools/test.py
innerlee: i see

tools/test.py
innerlee: Would higher value be better?

innerlee:
Notes:

I have changed the defaults to small values.
If one want more control, use `val_dataloader=dict(videos_per_gpu=1000)` or `test_dataloader=dict(videos_per_gpu=1000)`

innerlee (approved):

PR #261, opened 2020-10-18
[Fix] fix `EvalHook` when `key indicator` not in `eval metric`

dreamerlin:


mmaction/core/evaluation/eval_hooks.py
```diff
@@ -123,6 +124,11 @@ def evaluate(self, runner, results):
             runner.log_buffer.output[name] = val
         runner.log_buffer.ready = True
         if self.key_indicator is not None:
+            if self.key_indicator not in eval_res:
+                warnings.warn('The key indicator for evaluation is not '
```
innerlee, line 128: should this be a warning or an error?

dreamerlin: A warning ? because this misalignment will not influence the evaluation and the users still can find the best score manually from the json file.

itruonghai: where should I put key_indicator in config file. Is it like below
```
evaluation = dict(interval=1, metrics=['mean_average_precision'], key_indicator = 'mean_average_precision')
```

dreamerlin: > where should I put key_indicator in config file. Is it like below
> 
> ```
> evaluation = dict(interval=1, metrics=['mean_average_precision'], key_indicator = 'mean_average_precision')
> ```

Yep

itruonghai: It is auc not mean_average_precision, right?

dreamerlin: You can assign `rule='greater'` and `key_indicator='mean_average_precision'`.

innerlee (approved):

PR #262, opened 2020-10-19
WIP: [Feature]MultiModality: Text

SuX97:
+ Add text embedding dataset(called AuxiliaryFeatureDataset) and pipeline
+ Add text embedding preparation tools and k400 description classification result and fusion rgb result
+ Add self-supervised training (A re-implementation of this [paper](https://arxiv.org/pdf/2001.05691.pdf))

### DIFF
--- a/mmaction/datasets/pipelines/loading.py
+++ b/mmaction/datasets/pipelines/loading.py
@@ -1,4 +1,5 @@
 import io
+import json
 import os
 import os.path as osp
 import shutil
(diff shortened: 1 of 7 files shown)


tools/youtube_crawler.py
```diff
@@ -0,0 +1,47 @@
...
+    while True:
+        try:
+            page = requests.get(url).text
+            return page
+        except SSLError:
+            print(f'Connection error {url}.')
+            continue
+
+
+def get_title(page):
```
kennymckormick, line 22: Actually, you can use youtube-dl to get metadata like title or descriptions or subtitles. No need to write your own code.

SuX97: Right, I just missed that `--get-title` args.

val_titles.txt
```diff
@@ -0,0 +1,1800 @@
+zY_NOm7Qb8s LAGUNA PARK GUAYAQUIL ECUADOR SLACKLINE
```
dreamerlin, line 1: Is this file added by mistake?

SuX97: Yeah, will be removed later.

mmaction/datasets/pipelines/loading.py
```diff
@@ -955,6 +956,89 @@ def __call__(self, results):
...
+class LoadFrameLevelAuxiliaryFeature(object):
+    """Load frame level aux-feature.
+        Required keys are "featurepath" and "frame_inds",
+        added or modified keys are "features", "sampled_inds", "input_shape".
+        Args:
+            # FIXME
+    """
+
+    def __init__(self, raw_feature_ext='.json', feature_name='text'):
+        valid_raw_feature_ext = ('.json')
```
dreamerlin, line 969: ('.json', )

PR #266, opened 2020-10-21
[Feature] Add AVA dataset and transformation

dreamerlin:


kennymckormick:
@dreamerlin There is still conflicts need to be resolved

mmaction/datasets/ava_dataset.py
```diff
@@ -0,0 +1,241 @@
...
+                 timestamp_end=1798):
+        self.exclude_file = exclude_file
+        self.label_file = label_file
+        self.proposal_file = proposal_file
+        self.filename_tmpl = filename_tmpl
+        self.num_max_proposals = num_max_proposals
+        self.timestamp_start = timestamp_start
+        self.timestamp_end = timestamp_end
+        self.logger = get_root_logger()
+        super().__init__(
```
innerlee, line 100: why put super.init in the middle?

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -86,6 +86,288 @@ def __call__(self, results):
...
+        return results
+
+    def __repr__(self):
+        repr_str = (f'{self.__class__.__name__}('
+                    f'scales={self.scales}, mode={self.mode})')
+        return repr_str
+
+
+@PIPELINES.register_module()
+class BoxRescale:
```
innerlee, line 156: name proposal: `EntityBoxRescale`

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -86,6 +86,288 @@ def __call__(self, results):
...
+                scores = None
+            proposals = (proposals * img_scale).astype(np.float32)
+            results['proposals'] = proposals * scale_factor
+            results['scores'] = scores
+
+        return results
+
+
+@PIPELINES.register_module()
+class BoxCrop:
```
innerlee, line 195: name

innerlee:
looks good

mmaction/datasets/ava_dataset.py
dreamerlin: we should first assign some inner attributes for `super().__init`

innerlee: why not put them in the argument list?

dreamerlin: They are commonly not used for `BaseDataset`

innerlee: if they are not used, then why they are required to be assigned beforehand?

dreamerlin: https://github.com/open-mmlab/mmaction2/blob/master/mmaction/datasets/base.py#L61-L62, and we rewrite the `self.load_annotations()`.

innerlee: ok. add a line of comment explaining this

innerlee (approved):

PR #278, opened 2020-10-24
[Fix] fix data preparation bug

dreamerlin:


innerlee (approved):

innerlee:
scripts are too easy to be break. hope someone could add scripts to tests

yuhuixu1993:
Thanks for your contribution, I merged the updated tools/sthv2. It seems that the problem of sthv2 still exists.

dreamerlin:
> Thanks for your contribution, I merged the updated tools/sthv2. It seems that the problem of sthv2 still exists. @dreamerlin. Do I need to regenerate the raw frames?

Actually not. Is It the same problem? I test it and go with problems. BTW, it seems necessary to add unittest for the dataset preparation : )

yuhuixu1993:
@dreamerlin ,  I have solved it, thanks for your contribution!

dreamerlin:
> @dreamerlin , I have solved it, thanks for your contribution!

Glad to hear that !

PR #285, opened 2020-10-27
[Improve] Use softmax from 'scipy.special'

dreamerlin:


kennymckormick:
Seems we should not use those heavy packages like scipy in this repo, what's the motivation of this PR?

dreamerlin:
> Seems we should not use those heavy packages like scipy in this repo, what's the motivation of this PR?

So we should remove these following lines instead, right?
https://github.com/open-mmlab/mmaction2/blob/master/tools/analysis/report_accuracy.py#L4
https://github.com/open-mmlab/mmaction2/blob/master/tools/data/activitynet/activitynet_feature_postprocessing.py#L7

`scipy` has been used not in `tools/` though not in `mmaction2/mmaction` dir. 
But it has been recorded as [`known_third_party` ](https://github.com/open-mmlab/mmaction2/blob/master/setup.cfg#L22)

dreamerlin:
I vote for replace `scipy.interpolate.interp1d` and `scipy.special.softmax` with our customized function or other packages we used.

kennymckormick:
> > Seems we should not use those heavy packages like scipy in this repo, what's the motivation of this PR?
> 
> So we should remove these following lines instead, right?
> https://github.com/open-mmlab/mmaction2/blob/master/tools/analysis/report_accuracy.py#L4
> https://github.com/open-mmlab/mmaction2/blob/master/tools/data/activitynet/activitynet_feature_postprocessing.py#L7
> 
> `scipy` has been used not in `tools/` though not in `mmaction2/mmaction` dir.
> But it has been recorded as [`known_third_party` ](https://github.com/open-mmlab/mmaction2/blob/master/setup.cfg#L22)

You are right, seems the pre-commit hook only checks codes in the folder `mmaction`, so that the usage of `scipy` in tools is not tracked. I also think replacing with customized functions is a good idea. 
By the way, it is possible to let the pre-commit hook also check codes in folders like `tools`?

dreamerlin:
> > > Seems we should not use those heavy packages like scipy in this repo, what's the motivation of this PR?
> > 
> > 
> > So we should remove these following lines instead, right?
> > https://github.com/open-mmlab/mmaction2/blob/master/tools/analysis/report_accuracy.py#L4
> > https://github.com/open-mmlab/mmaction2/blob/master/tools/data/activitynet/activitynet_feature_postprocessing.py#L7
> > `scipy` has been used not in `tools/` though not in `mmaction2/mmaction` dir.
> > But it has been recorded as [`known_third_party` ](https://github.com/open-mmlab/mmaction2/blob/master/setup.cfg#L22)
> 
> You are right, seems the pre-commit hook only checks codes in the folder `mmaction`, so that the usage of `scipy` in tools is not tracked. I also think replacing with customized functions is a good idea.
> By the way, it is possible to let the pre-commit hook also check codes in folders like `tools`?

@hellock

PR #292, opened 2020-10-29
[Fix] Fix multi-node dist test

dreamerlin:
ref: https://github.com/open-mmlab/mmdetection/pull/4017

### DIFF
--- a/mmaction/apis/test.py
+++ b/mmaction/apis/test.py
@@ -107,7 +107,7 @@ def collect_results_cpu(result_part, size, tmpdir=None):
                                 dtype=torch.uint8,
                                 device='cuda')
         if rank == 0:
-            tmpdir = tempfile.mkdtemp()
+            tmpdir = tempfile.mkdtemp(dir='.dist_test')
             tmpdir = torch.tensor(
                 bytearray(tmpdir.encode()), dtype=torch.uint8, device='cuda')
             dir_tensor[:len(tmpdir)] = tmpdir
--- a/mmaction/apis/test.py
+++ b/mmaction/apis/test.py
@@ -116,8 +116,11 @@ def collect_results_cpu(result_part, size, tmpdir=None):
         tmpdir = dir_tensor.cpu().numpy().tobytes().decode().rstrip()
     else:
         mmcv.mkdir_or_exist(tmpdir)
+    # synchronizes all processes to make sure tmpdir exist
+    dist.barrier()
     # dump the part result to the dir
     mmcv.dump(result_part, osp.join(tmpdir, f'part_{rank}.pkl'))
+    # synchronizes all processes for loding pickle file
     dist.barrier()
     # collect all parts
     if rank != 0:
--- a/mmaction/apis/test.py
+++ b/mmaction/apis/test.py
@@ -8,6 +8,8 @@ import torch
 import torch.distributed as dist
 from mmcv.runner import get_dist_info
 
+from ..utils import get_random_string, get_shm_dir, get_thread_id
+
 
 def single_gpu_test(model, data_loader):
     """Test model with a single gpu.
@@ -107,7 +109,10 @@ def collect_results_cpu(result_part, size, tmpdir=None):
                                 dtype=torch.uint8,
                                 device='cuda')
         if rank == 0:
-            tmpdir = tempfile.mkdtemp(dir='.dist_test')
+            random_string = get_random_string()
+            thread_id = get_thread_id()
+            tmpdir = osp.join(get_shm_dir(), f'{random_string}_{thread_id}')
+            tmpdir = tempfile.mkdtemp(dir=tmpdir)
             tmpdir = torch.tensor(
                 bytearray(tmpdir.encode()), dtype=torch.uint8, device='cuda')
             dir_tensor[:len(tmpdir)] = tmpdir
--- a/mmaction/apis/test.py
+++ b/mmaction/apis/test.py
@@ -8,8 +8,6 @@ import torch
 import torch.distributed as dist
 from mmcv.runner import get_dist_info
 
-from ..utils import get_random_string, get_shm_dir, get_thread_id
-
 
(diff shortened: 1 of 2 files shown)


mmaction/apis/test.py
```diff
@@ -107,7 +107,7 @@ def collect_results_cpu(result_part, size, tmpdir=None):
                                 dtype=torch.uint8,
                                 device='cuda')
         if rank == 0:
-            tmpdir = tempfile.mkdtemp()
+            tmpdir = tempfile.mkdtemp(dir='.dist_test')
```
kennymckormick, line 110: That would be nice. BTW, I would like to mention that there is still risk in this version if you test multiple models at the same time, i.e. the middle results of different test runs will mess up. What about generating a hash based on the testing config and testing checkpoint, then use it as the name of the tmpdir.

dreamerlin: The case you mentioned is for `tmpdir is not None`, which is for Line#116

kennymckormick: I think, actually, the two cases are the same. If `tmpdir` is set, testing can run well when running on multiple nodes. But no user would like to designate `tmpdir` every time. If we can generate a tmpdir with hashing as its name each time, some efforts would be saved. Anyway, this PR is a good step.

kennymckormick (approved):

mmaction/apis/test.py
dreamerlin: How about use time-hash for tmpdir name, since it is difficult to obtain testing config and testing checkpoint here.

innerlee: > if you test multiple models at the same time

You mean testing multiple models in a same session?

innerlee: my understanding is that `mkdtemp` will create a unique folder name?

kennymckormick: > How about use time-hash for tmpdir name, since it is difficult to obtain testing config and testing checkpoint here.

I'm not sure, since we have multiple processes running at the same time, maybe they will get different hashes.

kennymckormick: > my understanding is that `mkdtemp` will create a unique folder name?

That's right. The problem is, by default it creates a tempdir under /tmp, while for different nodes on HPC, they all have their own /tmp.

kennymckormick: > > if you test multiple models at the same time
> 
> You mean testing multiple models in a same session?

At the same time & on the same storage system.

mmaction/apis/test.py
```diff
@@ -107,7 +109,10 @@ def collect_results_cpu(result_part, size, tmpdir=None):
                                 dtype=torch.uint8,
                                 device='cuda')
         if rank == 0:
-            tmpdir = tempfile.mkdtemp(dir='.dist_test')
+            random_string = get_random_string()
+            thread_id = get_thread_id()
+            tmpdir = osp.join(get_shm_dir(), f'{random_string}_{thread_id}')
+            tmpdir = tempfile.mkdtemp(dir=tmpdir)
             tmpdir = torch.tensor(
```
innerlee, line 112: dir_tensor= ？

dreamerlin: What are you confused about ?

innerlee: typo i guess

dreamerlin: `dir_tensor` is defined in Line#105

innerlee:
can this be tested?

dreamerlin:
> can this be tested?

Since the unittest coverage is down below 85%, we will check the codecov and raise an issue to some assigners to improve the unittest. And for this PR, it has been tested successfully ; )

innerlee:
does the test fail without this fix?

dreamerlin:
> does the test fail without this fix?

Actually, I haven't encount failure when performing multi-node dist test. This PR is related with open-mmlab/mmdetection#4017

innerlee:
yeah it's better to add  a test, so that we know what we have fixed.

innerlee:
@kennymckormick could explain more about the original problem, and why this is a partial fix?

mmaction/apis/test.py
hellock: `tempfile.mkdtemp(dir='.dist_test')` will create a temp dir with a random name under `.dist_test`, not under `/tmp`.

dreamerlin:
Unittest are added in https://github.com/open-mmlab/mmaction2/pull/322

innerlee:
I noticed that #322 has been merged, and no errors was thrown. So could u give details on what bug does this pr solve?

dreamerlin:
> I noticed that #322 has been merged, and no errors was thrown. So could u give details on what bug does this pr solve?

The bug will happen when performing distributed testing and gethering the result file under `tmp` directory. In some cases, the `tmp` directory will not be shared in multi-machines.

innerlee:
btw where is the `'.dist_test'`? under `mmaction/` folder?

I recommend save things in their workdir, so that we can make sure that
* the path is writable
* different runs wouldn't conflict

dreamerlin:
> btw where is the `'.dist_test'`? under `mmaction/` folder?
> 
> I recommend save things in their workdir, so that we can make sure that
> 
> * the path is writable
> * different runs wouldn't conflict

It is under `mmaction2` rather than `mmaction2/mmaction`

dreamerlin:
> btw where is the `'.dist_test'`? under `mmaction/` folder?
> 
> I recommend save things in their workdir, so that we can make sure that
> 
> * the path is writable
> * different runs wouldn't conflict

For `DistEpochEvalHook`, it will specify a `tmp_dir` using workdir (https://github.com/open-mmlab/mmaction2/blob/master/mmaction/core/evaluation/eval_hooks.py#L261-L265);
For testing, users can specify `arg.tmpdir` as the `tmp_dir`.

innerlee:
fine

PR #294, opened 2020-10-29
[Feature] Support Omnisource Dataset

kennymckormick:


tools/data/omnisource/trim_raw_video.py
```diff
@@ -0,0 +1,44 @@
...
+
+def trim(vid_name):
+    try:
+        lt = get_duration(vid_name)
+    except Exception:
+        print(f'get_duration failed for video {vid_name}', flush=True)
+        return
+
+    i = 0
+    name, suffix = vid_name.split('.')
```
innerlee, line 24: use splitext

tools/data/omnisource/trim_raw_video.py
```diff
@@ -0,0 +1,44 @@
...
+
+    command_tmpl = ('ffmpeg -y loglevel error -i {} -ss {} -t {} -crf 18 '
+                    '-c:v libx264 {}/part_{}.mp4')
+    while i * 10 < lt:
+        os.system(command_tmpl.format(vid_name, i * 10, 10, dest, i))
+        i += 1
+
+    # remove the raw video after decomposing it into 10-second clips to save
+    # space
+    os.system(f'rm {vid_name}')
```
innerlee, line 39: use `os.remove`

it is more safe, in case that someone name a video as `-rf ~`
okay joking

innerlee:
for paths and file operations, we prefer the dedicated functions

tools/data/omnisource/trim_raw_video.py
```diff
@@ -0,0 +1,44 @@
...
+        print(f'get_duration failed for video {vid_name}', flush=True)
+        return
+
+    i = 0
+    name, suffix = vid_name.split('.')
+
+    # We output 10-second clips into the folder `name`
+    dest = name
+    if not osp.exists(dest):
+        os.system('mkdir -p {}'.format(dest))
```
dreamerlin, line 29: `mmcv.mkdir_or_exist`

tools/data/omnisource/preparing_omnisource.md
```diff
@@ -0,0 +1,144 @@
...
+|   k200_train    |  76030   |  45.6G  |       N/A        |           N/A            |             N/A              |
+|    k200_val     |   4838   |  2.9G   |       N/A        |           N/A            |             N/A              |
+| googleimage_200 | 3050880  | 265.5G  |   TSN-R50-8seg   |         1188695          |             967              |
+|  insimage_200   | 3654650  | 224.4G  |   TSN-R50-8seg   |          879726          |             116              |
+|  insvideo_200   |  732855  | 1487.6G | SlowOnly-8x8-R50 |          330680          |             956              |
+| k200_raw_train  |  76027   | 963.5G  | SlowOnly-8x8-R50 |           N/A            |             N/A              |
+
+The file structure of our uploaded OmniSource dataset looks like:
+
+```
```
dreamerlin, line 29: better to align the explanation at the right from the same beginning.

tools/data/omnisource/preparing_omnisource.md
```diff
@@ -0,0 +1,144 @@
...
+    ├── vol_0.tar
+    ├── ...
+    └── vol_15.tar
+```
+
+## Data Preparation
+
+For data preparation, you need to first download those data. For `kinetics_200` and 3 web datasets: `googleimage_200`, `insimage_200` and `insvideo_200`, you just need to extract each volume and merge their contents. For Kinetics raw videos, since loading long videos is very heavy, you need to first trim it into clips. Here we provide a script named `trim_raw_video.py`. It trims a long video into 10-second clips and remove the original raw video. You can use it to trim the Kinetics raw video. The data should be placed in `data/OmniSource/`. When data preparation finished, the folder structure of `data/OmniSource` looks like (We omit the files not needed in training & testing for simplicity):
+
+```
```
dreamerlin, line 82: better to align the explanation at the right from the same beginning.

tools/data/omnisource/trim_raw_video.py
kennymckormick: u r right 🤣

tools/data/omnisource/preparing_omnisource.md
kennymckormick: Fixed, seems there is some bug with my editor.

tools/data/omnisource/preparing_omnisource.md
kennymckormick: Not quite sure what's the meaning. Now I split the long line into 3 paragraphs

dreamerlin: It is good now

innerlee (approved):

PR #304, opened 2020-11-01
[Feature] Support TAM [WIP]

congee524:


mmaction/models/backbones/tanet.py
```diff
@@ -0,0 +1,113 @@
...
+        num_segments (int): Number of frame segments.
+        tam_cfg (dict | None): Config for temporal adaptive module (TAM).
+            If set to None, it uses ```dict()``` to constrcut TAM.
+            Default: None.
+    """
+
+    def __init__(self, block, num_segments, tam_cfg=None):
+        super().__init__()
+        if tam_cfg is None:
+            tam_cfg = dict()
```
dreamerlin, line 27: Actually, the mutable value can be set as default values, as long as you use `copy.deepcopy()` to protect it. The method is also approved by @innerlee

PR #307, opened 2020-11-02
[Fix] Fix hvu object numclasses

kennymckormick:


innerlee:
any checkpoints need to be updated?

kennymckormick:
> any checkpoints need to be updated?

The ckpt of object baseline needs to be updated, will be uploaded by @dreamerlin . The change is just removing a class from weights of fc, no performance change.

innerlee:
Okay, waiting for the updated checkpoints

kennymckormick:
> Okay, waiting for the updated checkpoints

The checkpoint link has been already updated.

innerlee (approved):

Issue #306, opened 2020-11-01
SampleAVAFrames bug


irvingzhang0512:
Thank you very much for your work!
I'm looking forward to your ava models in Nov.

SampleAVAFrames works fine if `clip_len=32, frame_interval=2`, like in the unittest.
But when `clip_len=8, frame_interval=8`, [this function](https://github.com/open-mmlab/mmaction2/blob/master/mmaction/datasets/pipelines/loading.py#L419) only returns 2 index, but 8 index are expected.

My guess is that `length` is not calculated in a proper way. 
```python
        length = len(
            range(-self.frame_interval,
                  -(ori_clip_len + 1) // self.frame_interval, # should divide by 2 instead of self.frame_interval
                  -self.frame_interval))

        length = len(
            range(0, (ori_clip_len + 1) // self.frame_interval, # should divide by 2 instead of self.frame_interval
                  self.frame_interval))
```

kennymckormick:
Yes, -self.frame_interval should not be the stepsize and there is a bug in the current code. Actually, I think the current implementation of get_clips is too lengthy. I provide a new implementation in PR #317 (under review now). You can check that for reference.

PR #317, opened 2020-11-06
[Fix] Fix#306

kennymckormick:


innerlee:
You can try `git rebase` later, which produces a cleaner history.

Tests that capture this bug should be added.

dreamerlin:
Accordint to https://github.com/open-mmlab/mmaction2/issues/306, u can set another case `clip_len=8, frame_interval=8` in pytest : )

dreamerlin (approved):

PR #338, opened 2020-11-11
[Improve] Harmless changes according to Pylint

dreamerlin:


mmaction/apis/test.py
```diff
@@ -181,3 +180,4 @@ def collect_results_gpu(result_part, size):
         # the dataloader may pad some samples
         ordered_results = ordered_results[:size]
         return ordered_results
+    return None
```
innerlee, line 183: double check that None is a valid return value

dreamerlin: `None` is a valid return value by checking

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -252,10 +252,7 @@ def __init__(self, flip_ratio=0.5, direction='horizontal'):
         self.direction = direction
 
     def __call__(self, results):
-        if np.random.rand() < self.flip_ratio:
-            flip = True
-        else:
-            flip = False
+        flip = bool(np.random.rand() < self.flip_ratio)
```
innerlee, line 255: ~~bool~~

mmaction/models/backbones/x3d.py
```diff
@@ -391,7 +393,7 @@ def make_res_layer(self,
         if self.se_style == 'all':
             use_se = [True] * blocks
         elif self.se_style == 'half':
-            use_se = [True if i % 2 == 0 else False for i in range(blocks)]
+            use_se = [bool(i % 2 == 0) for i in range(blocks)]
```
innerlee, line 396: ~~bool~~

innerlee:
generally okay

innerlee (approved):

PR #341, opened 2020-11-11
[Fix] Fix docstring

dreamerlin:


mmaction/models/backbones/resnet3d.py
```diff
@@ -50,16 +47,15 @@ def __init__(self,
...
                  style='pytorch',
                  inflate=True,
-                 inflate_style='3x1x1',
                  non_local=False,
                  non_local_cfg=dict(),
                  conv_cfg=dict(type='Conv3d'),
                  norm_cfg=dict(type='BN3d'),
                  act_cfg=dict(type='ReLU'),
-                 with_cp=False):
+                 with_cp=False,
```
innerlee, line 55: Though not belonging to this pr, I do not like the `with_cp` name

mmaction/models/backbones/resnet3d.py
```diff
@@ -50,16 +47,15 @@ def __init__(self,
...
+                 **kwargs):
```
innerlee, line 56: where is this `kwargs` used?

dreamerlin: To absorb unused `inflate_style` argument

innerlee: then explicitly check it?

innerlee: nonono, only `inflate_style` is allowed in `kwargs`

dreamerlin: How about just keeping the `inflate_style` and mentioning that it is unused in `BasicBlock3d`. Otherwise it is wired to check the kwargs, since it contributes nothing for the class

innerlee: `set(kwargs.keys()).issubset(set(['inflate_style ']))`

we want to make sure that only inflate_style is passed into kwargs, so that typos like `activation_cfg` could be captured

innerlee: forgot assert

innerlee (approved):

PR #344, opened 2020-11-12
[Docs] Reorgnize docs

dreamerlin:

PR #345, opened 2020-11-12
[Feature] Add print_config scripts

dreamerlin:


dreamerlin:
Docs will be added in https://github.com/open-mmlab/mmaction2/pull/344

innerlee (approved):

PR #346, opened 2020-11-14
[Fix] Fix test bug

dreamerlin:


tools/test.py
```diff
@@ -116,7 +116,7 @@ def main():
         ('Please specify at least one operation (save or eval the '
          'results) with the argument "--out" or "--eval"')
 
-    if output_config:
+    if output_config.get('out', None):
```
innerlee, line 119: what if when output config is None?

dreamerlin: https://github.com/open-mmlab/mmaction2/blob/master/tools/test.py#L104

dreamerlin: I think it can not be `None`, due to https://github.com/open-mmlab/mmaction2/blob/master/tools/test.py#L104

innerlee (approved):

PR #354, opened 2020-11-17
[Fix] Fix `get_thread_id` when not using disk as backend

dreamerlin:


mmaction/datasets/pipelines/loading.py
```diff
@@ -901,7 +901,7 @@ def __call__(self, results):
 
 @PIPELINES.register_module()
 class OpenCVInit:
-    """Using OpenCV to initalize the video_reader.
+    """Using OpenCV to initialize the video_reader.
```
innerlee, line 904: 😕 This typo is been fixed many times...

dreamerlin: Got it! a stupid typo may occur in many places. It is a better strategy to global check, when finding a typo word.

innerlee (approved):

PR #359, opened 2020-11-19
[Improvement] Update docs about test crops.

irvingzhang0512:
## BUG
+ Describe bug: Try to run `tools/test.py` with TSM model and `DenseSampleFrames` with default [config](https://github.com/open-mmlab/mmaction2/blob/master/configs/recognition/tsm/tsm_r50_dense_1x1x8_100e_kinetics400_rgb.py) and [checkpoint](https://download.openmmlab.com/mmaction/recognition/tsm/tsm_r50_dense_1x1x8_100e_kinetics400_rgb/tsm_r50_dense_1x1x8_100e_kinetics400_rgb_20200626-91a54551.pth), but failed. 
+ Solution: Finally I find that, when using TSM models, we HAVE TO SET `test_cfg['test_crops']` MANUALLY to get correct results.
+ Thoughts: Maybe we should find a way to set `test_crops` automatically, not manually, like `Recognizer3D` models?

## TODO
+ [x] In order to let users to have better understanding about test crops, add more docs about this.
+ [x] Update TSM DenseSampleFrames default params.
+ [x] calculate num_segs for Recoginizer2D models automatically.
+ [x] changelog

### DIFF
--- a/configs/recognition/tsm/README.md
+++ b/configs/recognition/tsm/README.md
@@ -57,7 +57,12 @@ e.g., lr=0.01 for 4 GPUs * 2 video/gpu and lr=0.08 for 16 GPUs * 4 video/gpu.
 2. The **inference_time** is got by this [benchmark script](/tools/analysis/benchmark.py), where we use the sampling frames strategy of the test setting and only care about the model inference time,
 not including the IO time and pre-processing time. For each setting, we use 1 gpu and set batch size (videos per gpu) to 1 to calculate the inference time.
 3. The values in columns named after "reference" are the results got by training on the original repo, using the same model settings. The checkpoints for reference repo can be downloaded [here](https://download.openmmlab.com/mmaction/recognition/tsm/tsm_reference_ckpt.rar).
-4. There are two kinds of test settings for Something-Something dataset, efficient setting (center crop * 1 clip) and accurate setting (Three crop * 2 clip), which is referred from the [original repo](https://github.com/mit-han-lab/temporal-shift-module/tree/8d53d6fda40bea2f1b37a6095279c4b454d672bd).
+4. When using multiple crops for testing, we have to set `cfg.test_cfg['test_crops']` and `cfg.test_cfg['twice_sample']` manually to get correct results. Number of crops is calculated by:
+   1. `twice_sample` in `SampleFrames`
+   2. `num_sample_positions` in `DenseSampleFrames`
+   3. `ThreeCrop/TenCrop/MultiGroupCrop` in `test_pipeline`
+   4. `num_clips` in `SampleFrames` or its subclass if `clip_len != 1`
+5. There are two kinds of test settings for Something-Something dataset, efficient setting (center crop * 1 clip) and accurate setting (Three crop * 2 clip), which is referred from the [original repo](https://github.com/mit-han-lab/temporal-shift-module/tree/8d53d6fda40bea2f1b37a6095279c4b454d672bd).
 We use efficient setting as default provided in config files, and it can be changed to accurate setting by
 ```python
 ...
--- a/configs/recognition/tsm/README.md
+++ b/configs/recognition/tsm/README.md
@@ -57,17 +57,12 @@ e.g., lr=0.01 for 4 GPUs * 2 video/gpu and lr=0.08 for 16 GPUs * 4 video/gpu.
 2. The **inference_time** is got by this [benchmark script](/tools/analysis/benchmark.py), where we use the sampling frames strategy of the test setting and only care about the model inference time,
 not including the IO time and pre-processing time. For each setting, we use 1 gpu and set batch size (videos per gpu) to 1 to calculate the inference time.
 3. The values in columns named after "reference" are the results got by training on the original repo, using the same model settings. The checkpoints for reference repo can be downloaded [here](https://download.openmmlab.com/mmaction/recognition/tsm/tsm_reference_ckpt.rar).
-4. When using multiple crops for testing, we have to set `cfg.test_cfg['test_crops']` and `cfg.test_cfg['twice_sample']` manually to get correct results. Number of crops is calculated by:
-   1. `twice_sample` in `SampleFrames`
-   2. `num_sample_positions` in `DenseSampleFrames`
-   3. `ThreeCrop/TenCrop/MultiGroupCrop` in `test_pipeline`
-   4. `num_clips` in `SampleFrames` or its subclass if `clip_len != 1`
-5. There are two kinds of test settings for Something-Something dataset, efficient setting (center crop * 1 clip) and accurate setting (Three crop * 2 clip), which is referred from the [original repo](https://github.com/mit-han-lab/temporal-shift-module/tree/8d53d6fda40bea2f1b37a6095279c4b454d672bd).
+4. There are two kinds of test settings for Something-Something dataset, efficient setting (center crop * 1 clip) and accurate setting (Three crop * 2 clip), which is referred from the [original repo](https://github.com/mit-han-lab/temporal-shift-module/tree/8d53d6fda40bea2f1b37a6095279c4b454d672bd).
 We use efficient setting as default provided in config files, and it can be changed to accurate setting by
 ```python
 ...
 # `test_cfg = dict(average_clips=None)` for efficient setting
-test_cfg = dict(average_clips='prob', test_crops=3, twice_sample=True)  # for accurate setting
+test_cfg = dict(average_clips='prob')  # for accurate setting
 ...
 test_pipeline = [
     dict(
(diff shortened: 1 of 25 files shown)


configs/recognition/tsm/tsm_r50_dense_1x1x8_100e_kinetics400_rgb.py
```diff
@@ -18,7 +18,7 @@
         is_shift=True))
 # model training and testing settings
 train_cfg = None
-test_cfg = dict(average_clips=None)
+test_cfg = dict(average_clips='prob', test_crops=10)
```
dreamerlin, line 21: Since `DenseSampleFrames` is a strategy about samping, it not proper to directly set `test_crops=10`. I think it is better to add a `dense_sample` key for this case.

irvingzhang0512: Correct me if I miss something. 

`num_crops` is calculated by the following 4 params.
1. `twice_sample` in `SampleFrames`
2. `num_sample_positions` in `DenseSampleFrames`
3. `ThreeCrop/TenCrop/MultiGroupCrop` in `test_pipeline`
4. `num_clips` in `SampleFrames` or its subclass if `clip_len != 1`

So maybe in `test_cfg`, we should define 4 vars, corresponding to the above 4 params, which will make this more clearly for users.

irvingzhang0512: However, I think setting `test_crops` automatically may be a better solution

configs/recognition/tsm/README.md
```diff
@@ -62,7 +62,7 @@ We use efficient setting as default provided in config files, and it can be chan
 ```python
 ...
 # `test_cfg = dict(average_clips=None)` for efficient setting
-test_cfg = dict(average_clips='prob', test_crops=3, twice_sample=True)  # for accurate setting
+test_cfg = dict(average_clips='prob')  # for accurate setting
```
dreamerlin, line 65: It is for efficient setting. And acutally, the `average_clips='prob'` is the same with `average_clips=None`, since when the `crop=1` there is no need to ensemble.

dreamerlin: Ok, I got it, the modified code will automatically calculate the `test_crops`

mmaction/models/recognizers/recognizer2d.py
```diff
@@ -57,13 +54,27 @@ def forward_test(self, imgs):
...
+        #   4) `num_clips` in `SampleFrames` or its subclass if `clip_len != 1`
         cls_score = self.cls_head(x, num_segs)
-        if test_crops is not None:
-            if twice_sample:
-                test_crops = test_crops * 2
-            cls_score = self.average_clip(cls_score, test_crops)
 
-        return cls_score.cpu().numpy()
+        assert cls_score.size()[0] % batches == 0
+        if cls_score.size()[0] != batches:
```
dreamerlin, line 67: It seems that this `if` is unnecessary. Although `average_clip` has no effect when `cls_score.size()[0] == batches`, it can still generate cls_score after softmax when setting `prob`

dreamerlin (approved):

mmaction/datasets/pipelines/loading.py
```diff
@@ -323,7 +323,8 @@ class DenseSampleFrames(SampleFrames):
         sample_range (int): Total sample range for dense sample.
             Default: 64.
         num_sample_positions (int): Number of sample start positions, Which is
-            only used in test mode. Default: 10.
+            only used in test mode. Default: 10. That is to say, by default,
```
innerlee, line 326: > Number of sample start positions,

> at least 10 crops for one input sample

The first one reads like it samples in time axis, while the second seem like it is on spatial. Which is correct?

BTW the argument name is also not very clear

dreamerlin: it samples in time axis. it is indeed not accurate using `10 crops` here.

mmaction/models/heads/tsm_head.py
```diff
@@ -78,7 +78,8 @@ def forward(self, x, num_segments):
 
         Args:
             x (torch.Tensor): The input data.
-            num_segments (int): Number of frame segments. Default: 8.
+            num_segments (int): Useless in TSMHead.
```
innerlee, line 81: Could you provide more info, why this option exists here?

mmaction/models/recognizers/base.py
```diff
@@ -77,10 +77,11 @@ def average_clip(self, cls_score, num_segs=1):
 
         Using different averaging types ('score' or 'prob' or None,
         which defined in test_cfg) to computed the final averaged
-        class score.
+        class score. Only called in test mode.
 
         Args:
             cls_score (torch.Tensor): Class score to be averaged.
+            num_segs (int): Number of clips for each input sample.
```
innerlee, line 84: There is a mixture of `num_segments` and `num_segs`

mmaction/datasets/pipelines/loading.py
irvingzhang0512: Maybe `10 clips` is more accurate.

mmaction/models/heads/tsm_head.py
irvingzhang0512: In Recognizer2D, all [head_cls](https://github.com/open-mmlab/mmaction2/blob/master/mmaction/models/recognizers/recognizer2d.py#L60) need two input params.
By default, this `num_segments == num_clips * clip_len * num_crops`, which is useful in TSN models. However, in TSM models, the `self.num_segments` we need is a required hyper-param to build models

mmaction/models/recognizers/base.py
irvingzhang0512: Maybe you mean a mixture of `clips` and `segs`? In addition to the mixture in this line, there has been a mixture of `clip` and `segs` in method name `average_clip` and function argument `num_segs`. 

I do understand this `num_segs` means test clips (both temporal samples from `twice_sample`/`num_clips` if `clip_len!=1`/`num_sample_positions ` and spatial samples from `ThreeCrop/TenCrop/MultiGroupCrop`) for one input sample.

I plan to modify this line to 
```
num_segs (int): Number of segments for each input sample.  Number of segments is calculated by: 
    1)twice_sample... 
    2) num_clips... 
    3) ThreeCrop... 
    4) num_sample_positions...
```

@dreamerlin Any suggestions?

dreamerlin: I think he means the mixture of variable name between `num_segments` and `num_segs`. we can unify it.

irvingzhang0512: ... sure

irvingzhang0512: > I think he means the mixture of variable name between `num_segments` and `num_segs`. we can unify it.

Actually in this project, there are a lot of `num_segs` and `num_segments`. 

As far as I'm concerned, `num_segments` is more like a model definition related param. So, in `TSMHead` and `average_clip`, `num_segs` is better, just like `TSNHead/TPNHead`. 

Is this ok for you? @dreamerlin

mmaction/datasets/pipelines/loading.py
innerlee: ya

mmaction/models/recognizers/base.py
dreamerlin: Agree. I think, it is hard to cause ambiguity for these two words.

irvingzhang0512:
@innerlee Could you please review this PR?

innerlee:
will take a look in the afternoon

mmaction/models/heads/tsm_head.py
```diff
@@ -73,13 +73,16 @@ def init_weights(self):
...
 
-    def forward(self, x, num_segments):
+    def forward(self, x, num_segs):
         """Defines the computation performed at every call.
 
         Args:
             x (torch.Tensor): The input data.
-            num_segments (int): Number of frame segments. Default: 8.
-
+            num_segs (int): Useless in TSMHead. By default, `num_segs`
```
innerlee, line 81: Whether this could be removed, if it is useless? May discuss later.

innerlee (approved):

mmaction/models/heads/tsm_head.py
irvingzhang0512: This argment could be removed if we add some codes in `recognizer2d.py`, codes like
```
if isinstance(self.cls_head, TSMHead):
    cls_score = self.cls_head(x)
else:
    cls_score = self.cls_head(x, num_segs)
```
Is this necessary? @innerlee 
Any better ideas? @dreamerlin

dreamerlin: I vote for keeping it unchanged and describe it in docstring as before

PR #379, opened 2020-11-25
Fix potential issue in long_video_demo.py

dreamerlin:
This PR fix 2 potential issue in long_video_demo.py
1. The last of some videos will be decoded as None in OpenCV. When it is the case, demo will skip it.
2. Popleft only once for keeping img shape.

### DIFF
--- a/demo/long_video_demo.py
+++ b/demo/long_video_demo.py
@@ -1,6 +1,5 @@
 import argparse
 import random
-import warnings
 from collections import deque
 from operator import itemgetter
 
@@ -72,7 +71,7 @@ def show_results():
         prog_bar.update()
         ret, frame = cap.read()
         if frame is None:
-            warnings.warn('A certain frame of the video is None.')
+            # drop it when encounting None
             continue
         backup_frames.append(np.array(frame)[:, :, ::-1])
         if ind == sample_length:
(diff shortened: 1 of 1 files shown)


demo/long_video_demo.py
```diff
@@ -70,6 +71,9 @@ def show_results():
         ind += 1
         prog_bar.update()
         ret, frame = cap.read()
+        if frame is None and ind == num_frames:
```
innerlee, line 74: why not just drop it?

innerlee: you know, things in videos are always inexact

demo/long_video_demo.py
```diff
@@ -71,9 +71,9 @@ def show_results():
         ind += 1
         prog_bar.update()
         ret, frame = cap.read()
-        if frame is None and ind == num_frames:
-            warnings.warn('The last frame of the video is None.')
-            break
+        if frame is None:
+            warnings.warn('A certain frame of the video is None.')
```
innerlee, line 75: no warning needed

demo/long_video_demo.py
dreamerlin: Indeed

innerlee (approved):

PR #381, opened 2020-11-25
[Feature] ACSampler

SuX97:
This PR adds ACSampler from the paper: *SCSampler*, including:
+ AC(action classification) samplers based on I-frame, Motion vector, audio.
+ I3D result improvement by using all three samplers.

This PR is challenging, as it is the first model that has a complicated pipeline operated on four kinds of modalities. 

ACSampler is the 'simplified' version introduced in the paper, that trains a sampler offline and can be adapted to all kinds of 3D models. As for  'SCSampler'(which requires joint training the recognizer along with the sampler), it will be addressed after ACSampler is re-implemented.


mmaction/datasets/audio_visual_dataset.py
```diff
@@ -0,0 +1,73 @@
...
+        with open(self.ann_file, 'r') as fin:
+            for line in fin:
+                line_split = line.strip().split()
+                video_info = {}
+                idx = 0
+                # idx for frame_dir
+                frame_dir = line_split[idx]
+                if self.audio_prefix is not None:
+                    audio_path = osp.join(self.audio_prefix,
+                                          frame_dir) + '.npy'
```
dreamerlin, line 43:
```
audio_path = osp.join(self.audio_prefix, frame_dir + '.npy')
```

Issue #374, opened 2020-11-24
Kinetics class rename problem


ShoufaChen:
Hi,

In [rename_classnames.sh](https://github.com/open-mmlab/mmaction2/blob/master/tools/data/kinetics/rename_classnames.sh), the `(` and ``)` will be converted to `-`.

However, the [parse_file_list.py](https://github.com/open-mmlab/mmaction2/blob/master/tools/data/parse_file_list.py#L320) only converts whitespace to '_', without changing the `(` .

This will cause mismatch at build_file_list.py:
```
        rgb_list, flow_list = list(), list()
        for item in split:
            if item[0] not in frame_info:
                continue
```

And some classes will be ignored, such as `skiing_(not_slalom_or_crosscountry)`.

I'm glad to pull a request if it's a bug. And please tell me if I misunderstood it.

kennymckormick:
Thanks a lot for reporting that bug. It will help a lot if you can provide a fix for this bug.

ShoufaChen:
I have a pr https://github.com/open-mmlab/mmaction2/pull/376 for this issue.

innerlee:
@dreamerlin whose bug? `parse_file_list.py` or those rename scripts?

ShoufaChen:
I think it's better to modify the rename script because other codebases, such as SlowFast, Non-Local also leave the `(` unchanged.

kennymckormick:
> @dreamerlin whose bug? `parse_file_list.py` or those rename scripts?

The bug is because the rename scripts and `parse_file_list.py` are not consistent. (related to Kinetics only)

kennymckormick:
@dreamerlin . I also find that it seems the renaming step should be compulsory, which is inconsistent with the description in README. Would u please help check if this step can be omitted?

dreamerlin:
> @dreamerlin . I also find that it seems the renaming step should be compulsory, which is inconsistent with the description in README. Would u please help check if this step can be omitted?

The step is indeed compulsory I think, but the docs doesn't point it out explicitly. I modify it to be more clear,

PR #384, opened 2020-11-26
fix rename kinetics classnames bug

ShoufaChen:
Previous `rename_classnames.sh` and `detox` can rename the
`(` and `)` to `-`, which is not desirable. [#374](https://github.com/open-mmlab/mmaction2/issues/374#issue-749310860)


dreamerlin:
Thank you very much !

dreamerlin (approved):

innerlee:
`detox` does much more than replacing whitespaces. Since here we are dealing with class names, which are very clean by themselves, I think this fix is okay.

For detoxing general filenames, one need to be more careful though.

PR #395, opened 2020-11-28
[Refactor] refactor EvalHook

dreamerlin:
## OverView

This PR refactors EvalHook

1. Add by_epoch param to determine perform evaluation by epoch or by iteration.
2. Add automatically assign a compare key when users do not set one.
3. Save checkpoint by creating a symlink as well as record best info in runner hook meta
4. Refactor the code structure

## Migration

Configs files need to be changed

1. Rename `EpochEvalHook` to `EvalHook`, `DistEpochEvalHook` to `DistEvalHook`.
2. Rename `key_indicator` to `save_best`.

For example, in `slowfast_kinetics_pretrained_r50_4x16x1_20e_ava_rgb.py` cfg file

from 

```python
evaluation = dict(interval=1, key_indicator='mAP@0.5IOU')
```

to

```python
evaluation = dict(interval=1, save_best='mAP@0.5IOU')
```

## Misc.
This Part may further be merged into `MMCV` or applied to other MM codebases.

### DIFF
--- a/mmaction/core/evaluation/eval_hooks.py
+++ b/mmaction/core/evaluation/eval_hooks.py
@@ -332,3 +332,21 @@ class DistEvalHook(EvalHook):
 
             if self.save_best:
                 self._save_ckpt(runner, key_score)
+
+
+class EpochEvalHook(EvalHook):
+    """Deprecated class for ``EvalHook``."""
+
+    def __init__(self, *args, **kwargs):
+        warnings.warn('"EpochEvalHook" is deprecated, please switch to'
+                      '"EvalHook"')
+        super().__init__(*args, **kwargs)
+
+
+class DistEpochEvalHook(DistEvalHook):
+    """Deprecated class for ``DistEvalHook``."""
+
+    def __init__(self, *args, **kwargs):
+        warnings.warn('"DistEpochEvalHook" is deprecated, please switch to'
+                      '"DistEvalHook"')
+        super().__init__(*args, **kwargs)
--- a/mmaction/core/evaluation/eval_hooks.py
+++ b/mmaction/core/evaluation/eval_hooks.py
@@ -147,11 +147,13 @@ class EvalHook(Hook):
 
     def after_train_iter(self, runner):
         """Called after every training iter to evaluate the results."""
-        self._do_evaluate(runner)
+        if not self.by_epoch:
+            self._do_evaluate(runner)
 
     def after_train_epoch(self, runner):
         """Called after every training epoch to evaluate the results."""
-        self._do_evaluate(runner)
+        if self.by_epoch:
+            self._do_evaluate(runner)
 
     def _do_evaluate(self, runner):
         """perform evaluation and save ckpt."""
--- a/mmaction/core/evaluation/eval_hooks.py
+++ b/mmaction/core/evaluation/eval_hooks.py
@@ -200,16 +200,16 @@ class EvalHook(Hook):
     def _save_ckpt(self, runner, key_score):
         if self.by_epoch:
             current = f'epoch_{runner.epoch + 1}'
+            cur_type, cur_time = 'epoch', runner.epoch + 1
         else:
             current = f'iter_{runner.iter + 1}'
+            cur_type, cur_time = 'iter', runner.iter + 1
 
         best_score = runner.meta['hook_msgs'].get(
             'best_score', self.init_value_map[self.rule])
         if self.compare_func(key_score, best_score):
             best_score = key_score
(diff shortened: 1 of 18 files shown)


mmaction/core/evaluation/eval_hooks.py
```diff
@@ -25,20 +23,26 @@ class EpochEvalHook(Hook):
...
             ``mean_average_precision``, ``mmit_mean_average_precision``
             for action recognition dataset (RawframeDataset and VideoDataset).
             ``AR@AN``, ``auc`` for action localization dataset
-            (ActivityNetDataset). Default: `top1_acc`.
-        rule (str | None): Comparison rule for best score. Options are None,
-            'greater' and 'less'. If set to None, it will infer a reasonable
-            rule. Default: 'None'.
+            (ActivityNetDataset). If ``save_best`` is ``auto``, the first key
+             of the returned ``OrderedDict`` result will be used. The interval
+             of ``CheckpointHook`` should device EvalHook. Default: None.
```
irvingzhang0512, line 40: what does `device EvalHook` mean? is this a typo?

Actually, I did some digging about hooks in mmcv. The best ckpt saved in `EvalHook` is a symlink of `runner.meta['hook_msgs']['last_ckpt']`, which is generated by `CheckpointHook`. So if checkpoint hook interval is 5 & eval hook interval is 1 & best ckpt is from epoch no.7, then the best ckpt we saved is from epoch no. 5.

Is this the case? Can we just save the ckpt in EvalHook?

dreamerlin: Save the checkpoint if the symlink can not be created ? @innerlee @irvingzhang0512

irvingzhang0512: I prefer saving ckpt instead of creating symlink. Since we don't have early stopping by default, if we set `max_keep_ckpts` in checkpoint hook, the best ckpt we save may be removed.

By the way, it seems that current codes doesn't remove previous best ckpt. Say we train for 10 epochs, and val accuracy increases epoch by epoch. By using our eval hook, 10 best ckpt symlink left..

dreamerlin: use cp instead of symlink now. Thanks!

mmaction/core/evaluation/eval_hooks.py
```diff
@@ -25,22 +23,33 @@ class EpochEvalHook(Hook):
...
             (AVADataset). Default: `top1_acc`.
-        rule (str | None): Comparison rule for best score. Options are None,
-            'greater' and 'less'. If set to None, it will infer a reasonable
-            rule. Default: 'None'.
+=======
+            ``AR@AN``, ``auc`` for action localization dataset
+            (ActivityNetDataset). If ``save_best`` is ``auto``, the first key
+             of the returned ``OrderedDict`` result will be used. The interval
+             of ``CheckpointHook`` should device EvalHook. Default: None.
+>>>>>>> polish again
```
congee524, line 47: is this what `git base` produced?

mmaction/core/evaluation/eval_hooks.py
```diff
@@ -222,8 +291,8 @@ def __init__(self,
                  dataloader,
                  start=None,
                  interval=1,
-                 save_best=True,
-                 key_indicator='top1_acc',
+                 by_epoch=True,
+                 save_best=None,
```
congee524, line 295: maybe we should rename `save_best`, since it represents the metrics for the best pth.

and assert whether it in the supported measure type.

dreamerlin: `save_best` serves as both boolean and str, it can provide whether to save best and save which best. Besides, it can be asserted in `_init_rule` as `key_indicator`.

innerlee:
Please help with user migration.

* Add a stub for the old classes `DistEpochEvalHook, EpochEvalHook`, and give proper error messages (ideally, also give links to this pr) to guide the user to update configs/codes.
* Give detailed migration guide on the top post in this thread. Like I commented at https://github.com/open-mmlab/mmaction2/pull/567#issuecomment-768080043

dreamerlin:
> Please help with user migration.
> 
> * Add a stub for the old classes `DistEpochEvalHook, EpochEvalHook`, and give proper error messages (ideally, also give links to this pr) to guide the user to update configs/codes.
> * Give detailed migration guide on the top post in this thread. Like I commented at [#567 (comment)](https://github.com/open-mmlab/mmaction2/pull/567#issuecomment-768080043)

Done

innerlee (approved):

PR #397, opened 2020-11-29
update mmcv version

dreamerlin:


innerlee:
How about < 1.3

mmaction/__init__.py
```diff
@@ -4,7 +4,7 @@
 from .version import __version__
 
 mmcv_minimum_version = '1.1.1'
-mmcv_maximum_version = '1.2.0'
+mmcv_maximum_version = '1.3.0'
 mmcv_version = digit_version(mmcv.__version__)
 
 assert (digit_version(mmcv_minimum_version) <= mmcv_version
```
innerlee, line 10: <

mmaction/__init__.py
```diff
@@ -4,7 +4,7 @@
 from .version import __version__
 
 mmcv_minimum_version = '1.1.1'
-mmcv_maximum_version = '1.2.0'
+mmcv_maximum_version = '1.3.0'
```
hellock, line 7: not 1.3.0 but 1.3

1.3rc0 should also be excluded.

mmaction/__init__.py
hellock: <= is okay

dreamerlin: I think `<=` is more reasonable, isn't it ?

innerlee: Minor version is breaking, isn't it?

hellock: `<= 1.3` means any version of 1.3 is excluded, such as 1.3.0, 1.3rc0, etc.

innerlee: If you allow 1.3, why not simply bound on v2?

hellock: 1.3.0 is not allowed.

hellock: `digit_version('1.3rc0') < digit_version('1.3.0')` will return True, but we assume that 1.3rc0 also breaks the compatibility.

innerlee: So lets bound on 1.3rc0?

innerlee: okay 1.3

innerlee (approved):

skyqwe123:
I don't know if the mmcv-full has been modified. It can be installed normally before, but the following error will be displayed after re-installation:
`(open-mmlab) user@user-ubuntu:/data6$ pip install mmcv-full
Collecting mmcv-full
  Using cached https://files.pythonhosted.org/packages/d7/ca/fa9f0aedeb6c2f701996a71ef03e365c2870b1bb0ea63f07459b1ef9128e/mmcv-full-1.2.1.tar.gz
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-0cfwdsnl/mmcv-full/setup.py", line 130
        f'OpenCV >=3 is required but {cv2.__version__} is installed')
                                                                   ^
    SyntaxError: invalid syntax
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-0cfwdsnl/mmcv-full/`

I've tried it several times, but there's always this error （the dependent package is normal）

PR #404, opened 2020-11-30
[Docs] Update readthedocs with audio recognition models

dreamerlin:
image


innerlee:
How about making recognition with audio under a new section called multi-modality?

innerlee:
hmm let's do this in later version.

innerlee (approved):

PR #445, opened 2020-12-15
Polish modelzoo statistics

congee524:
### theme
polish modelzoo statics

### content
add hyperlinks of the models and tables in Model Zoo Statistic page

### others
you can view the page in the [screencapture](https://drive.google.com/file/d/1LzFlk5E1c0PbKyBdmvOPARl6mIY_EcvJ/view?usp=sharing)


congee524:
Sphinx has some problems in rendering the markdown table.

In many cases, the table or the unordered list will be recognized as `code`, so the statistics of ckpt, papers and configs have been changed into paragraphs.

docs/merge_docs.sh
```diff
@@ -33,8 +33,8 @@ sed -i "s/md###t/html#t/g" demo.md
 sed -i 's/# Preparing/# /g' prepare_data.md
 sed -i 's/#/##&/' prepare_data.md
 
-sed -i '1i\## Action Localization Models' localization_models.md
-sed -i '1i\## Action Recognition Models' recognition_models.md
+sed -i '1i\# Action Localization Models' localization_models.md
```
innerlee, line 36: why change this line

congee524: I put the contents of recognition_models and localization_models to model zoo.md, so that it can jump inside the page

docs/stat.py
```diff
@@ -48,12 +90,19 @@
 modelzoo = f"""
 # Model Zoo Statistics
 
-* Number of checkpoints: {len(allckpts)}
-* Number of configs: {len(allconfigs)}
-* Number of papers: {len(allpapers)}
+Number of checkpoints: {len(allckpts)}</br>
```
innerlee, line 93: why change these

congee524: A models list is to be placed below these three lines of statistics and each model has a paperlist

Sphinx has some problems in rendering tables and unordered lists. Specifically,
If we set the model list to level 2 unordered list or higher, then after inserting a table, the remaining level 2 unordered list will be rendered as `code style`.
If we indent the table code, the table will be rendered as `code style`.

So I think I must set the model list as level 1 unordered list and thus I didn't set the three statistics as unordered list.

dreamerlin:
How is it going now ?

dreamerlin:
Kindly ping

kennymckormick:
@congee524 conflicts exist now.

innerlee:
I will make a new pr later, closing this for now

PR #446, opened 2020-12-15
Fix typo in BaseHead default loss_factor.

SuX97:


innerlee:
The problem is deeper. There must be some `**kwargs` not being used and also not being checked

SuX97:
> The problem is deeper. There must be some `**kwargs` not being used and also not being checked

There is no **kwargs used in building heads and losses. The reason why we haven't tested it and not error being revealed is that, in BaseWeightedLoss, loss_weight is set to be 1.0 by default. And in all our usages(e.g. all configs), we override the `loss_cls` by `dict(type='XXXLoss') ` without specifying loss_weight.

innerlee:
Is `loss_factor` a valid argument?

SuX97:
> Is `loss_factor` a valid argument?

Yes, but it will always be overridden together with the total dictionary of `loss_cls`. So we never met this error before.

tests/test_models/test_head.py
```diff
@@ -6,7 +6,7 @@
 
 
 class ExampleHead(BaseHead):
-    # use a ExampleHead to success BaseHead
+    # use a ExampleHead to succeed BaseHead
```
innerlee, line 9: What does `succeed BaseHead` mean?

SuX97: I am not sure, I believe the former author means something like 'inherit'. So I change it to the verb form.

innerlee: succeed -> test

innerlee (approved):

PR #447, opened 2020-12-15
[Refactor] Refactor metrics and localization utils structure

dreamerlin:
This PR refactors our metrics into a registry and adjusts the file structure about metrics and localization. For more details, 

1. Design a `build_metrics` function for common evaluation
2. Move some functions about `mean_ap`, `recall` and `iou` to corresponding files
3. Remove `mmaction/localizaton`, and move its parts into the corresponding ones in `mmaction/models/localizers/utils/`.


mmaction/core/evaluation/mean_ap.py
```diff
@@ -0,0 +1,251 @@
...
+        precision, recall, _ = binary_precision_recall_curve(score, label)
+        ap = -np.sum(np.diff(recall) * np.array(precision)[:-1])
+        results.append(ap)
+    results = [x for x in results if not np.isnan(x)]
+    if results == []:
+        return np.nan
+    return np.mean(results)
+
+
+# TODO: check the name
```
dreamerlin, line 165: Is the name of this function correct ? @kennymckormick

mmaction/core/evaluation/eval_detection.py
```diff
@@ -147,6 +148,8 @@ def evaluate(self):
         return self.mAP, self.average_mAP
 
 
+# TODO: This function seems similar
+#  with `def average_precision_at_temporal_iou`
```
dreamerlin, line 152: @kennymckormick

kennymckormick: The two functions perform nearly the same operation and can be merged as one.

mmaction/core/evaluation/mean_ap.py
kennymckormick: I think the name is not proper and should be changed

mmaction/core/evaluation/__init__.py
```diff
@@ -1,17 +1,23 @@
-from .accuracy import (average_precision_at_temporal_iou,
-                       average_recall_at_avg_proposals, confusion_matrix,
-                       get_weighted_score, interpolated_precision_recall,
-                       mean_average_precision, mean_class_accuracy,
-                       mmit_mean_average_precision, pairwise_temporal_iou,
-                       softmax, top_k_accuracy)
+from .accuracy import (MeanClsAcc, TopKAcc, confusion_matrix,
+                       get_weighted_score, mean_class_accuracy, top_k_accuracy)
```
kennymckormick, line 2: Seems TopKAcc and top_k_accuracy are duplicated now, while there is no way for a user to use top_k_accuracy

kennymckormick (commented):
After refactoring metrics, you need also refactor tools/test.py, remember that top_k_accuracy and mean_class_accuracy are not valid metrics now.

mmaction/core/evaluation/__init__.py
dreamerlin: `TopKAcc` with `top_k_accuracy` is similar with `BinaryLogisticRegressionLoss` with `binary_logistic_regression_loss`. If users want to use the function, they can directly import by `from mmaction.core import top_k_accuracy`.

Maybe I didn't get your meaning

mmaction/core/evaluation/eval_detection.py
dreamerlin: I am not familiar with these two, could you help to merge ?

mmaction/core/evaluation/mean_ap.py
dreamerlin: How about `interpolated_average_precision`

mmaction/core/evaluation/__init__.py
kennymckormick: I mean, we used to set `--eval top_k_accuracy` for testing. Now that we refactored the metrics, some docs may need to be changed (`--eval top_k_accuracy`  doesn't work now, since `top_k_accuracy` is not in allowed_metrics). 
And part of my last comments are not right: top_k_accuracy and TopKAcc are not duplicated.

mmaction/core/evaluation/eval_detection.py
kennymckormick: OK.

mmaction/core/evaluation/mean_ap.py
kennymckormick: That would be fine

mmaction/core/evaluation/__init__.py
dreamerlin: Maybe we can keep some old command like `--eval top_k_accuracy` for several versions but raise a deprecated warning until we totally remove them.

kennymckormick: Right, so you may need to modify some codes in evaluate functions of datasets

kennymckormick:
Have time to push this PR forward? I think it still doesn't support `--eval top_k_accuracy` now

PR #461, opened 2020-12-18
data_preprocess

wjn922:
fix some errors in download_video.sh
for activitynet, download.py relies on mmcv.
for gym, the virtual env name should be gym.

### DIFF
--- a/tools/data/activitynet/download_videos.sh
+++ b/tools/data/activitynet/download_videos.sh
@@ -4,6 +4,7 @@
 conda env create -f environment.yml
 source activate activitynet
 pip install --upgrade youtube-dl
+pip install mmcv
 
 DATA_DIR="../../../data/ActivityNet"
 python download.py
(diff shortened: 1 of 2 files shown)


innerlee:
Thanks for your contribution!

tools/data/activitynet/download_videos.sh
```diff
@@ -4,6 +4,7 @@
 conda env create -f environment.yml
 source activate activitynet
 pip install --upgrade youtube-dl
+pip install mmcv
```
innerlee, line 7: How about adding mmcv to env.yml

wjn922: That's also fine.

innerlee (approved):

tools/data/activitynet/download_videos.sh
innerlee: Yeah its not important

PR #310, opened 2020-11-03
Remove redundant code

kennymckormick:

PR #463, opened 2020-12-19
[Fix] Fix potential bug about output_config overwrite

CarolinaCheng616:
This PR fixes potential bug when users do not assign parser args. For example, if a user does not assign `args.out` 

```python
output_config = dict(out=f'{work_dir}/results.json', output_format='json')
args.out = None
output_config = Config._merge_a_into_b(dict(out=args.out), output_config)
```
then, `output_config` in config files will be overwritten.


dreamerlin (approved):

innerlee:
Thanks! This bug was introduced in https://github.com/open-mmlab/mmaction2/pull/310, in which we replaced our own merge function with mmcv's.

In our version of merge, we treat `None` as transparent, and would not overwrite things.
It was my oversight that, I wrongly thought mmcv's merge would respect this semantic.

This has caused many problems. Now I'm thinking whether we should bring our merge function back. mmcv's functions sometimes are less thoughtful :/

PR #468, opened 2020-12-21
[Feature] Sparsedemo

kennymckormick:
Add sparse demo:
- By setting the new arg `stride`, you can get predictions for every `stride * sample_size` frames.
- A feature required by web demo since the original one is too slow.


innerlee:
If I remembered correctly, the current demo implements an adaptive "stride". Conceptually there are two processes, 
* one process reads the video and write the frames on the screen, regardless whether prediction is finished or not.
* one process fetch one current frame, and do prediction, and write results on screen, and fetch frame again, looping

@dreamerlin

dreamerlin:
Is it the same with `INPUT_STEP` ?

demo/README.md
```diff
@@ -225,6 +225,7 @@ Optional arguments:
 - `INPUT_STEP`: Input step for sampling frames, which can help to get more spare input. If not specified , it will be set to 1.
```
dreamerlin, line 225: Here

kennymckormick: Nope, they are not the same. For example, if set `input_step` as 4, for every consecutive 4 frames, we randomly append one frame to the frame queue and drop the other three frames. However, getting predictions every 4 frames means you first get the prediction for frame 0-64, the 4-68, etc. 
The input-size might works for TSN, but not a general solution to both 2D and 3D approaches.

kennymckormick:
> If I remembered correctly, the current demo implements an adaptive "stride". Conceptually there are two processes,
> 
> * one process reads the video and write the frames on the screen, regardless whether prediction is finished or not.
> * one process fetch one current frame, and do prediction, and write results on screen, and fetch frame again, looping
> 
> @dreamerlin

Plz check my comments to @dreamerlin

dreamerlin:
> If I remembered correctly, the current demo implements an adaptive "stride". Conceptually there are two processes,
> 
> * one process reads the video and write the frames on the screen, regardless whether prediction is finished or not.
> * one process fetch one current frame, and do prediction, and write results on screen, and fetch frame again, looping
> 
> @dreamerlin

It is for camera demo, not for this one.

demo/long_video_demo.py
```diff
@@ -88,6 +95,11 @@ def show_results():
 
         ret, scores = inference()
 
+        if ret and stride > 0:
+            pred_stride = int(sample_length * stride)
+            for i in range(pred_stride):
+                frame_queue.popleft()
```
dreamerlin, line 101: Got! I think it is better to move this to `def inference` to avoid repeat popleft 

```python
def inference():
    if len(frame_queue) != sample_length:
        # Do no inference when there is no enough frames
        return False, None

    cur_windows = list(np.array(frame_queue))
    img = frame_queue.popleft()

    if stride >0:
        pred_stride = max(1, int(sample_length * stride))
        for _ in range(pred_stride - 1):
            frame_queue.popleft()

    ...
```
Better to rerun for checking.

kennymckormick: OK. Besides, I would like to remove global args.

demo/long_video_demo.py
```diff
@@ -120,38 +130,37 @@ def show_results():
...
     if next(model.parameters()).is_cuda:
-        cur_data = scatter(cur_data, [device])[0]
+        cur_data = scatter(cur_data, [args.device])[0]
     with torch.no_grad():
         scores = model(return_loss=False, **cur_data)[0]
+
+    if args.stride > 0:
+        pred_stride = int(args.sample_length * args.stride)
+        for i in range(pred_stride):
+            frame_queue.popleft()
```
dreamerlin, line 154: it seems not right. If `args.stride = 0` (not assigned),   `frame_queue` will not be popped ?

dreamerlin: OK, if the `deque` is full, it will automatically pop : )

dreamerlin (approved):

dreamerlin:
@innerlee cc.

demo/README.md
```diff
@@ -225,6 +225,7 @@ Optional arguments:
...
 - `DEVICE_TYPE`: Type of device to run the demo. Allowed values are cuda device like `cuda:0` or `cpu`. If not specified, it will be set to `cuda:0`.
 - `THRESHOLD`: Threshold of prediction score for action recognition. Only label with score higher than the threshold will be shown. If not specified, it will be set to 0.01.
+- `STRIDE`: By default, the demo generates a prediction for each single frame, which might cost lots of time. To speed up, you can set the argument `STRIDE` and then the demo will generate a prediction every `STRIDE x sample_length` frames. For example, if the sample_length is 64 frames and you set `STRIDE` to 0.5, predictions will be generated every 32 frames. If set as 0, predictions will be generated for each frame. Default: 0.
```
innerlee, line 228: What is `sample_length`? They are not clear in this page

demo/long_video_demo.py
```diff
@@ -45,12 +45,22 @@ def parse_args():
         type=float,
         default=0.01,
         help='recognition score threshold')
+    parser.add_argument(
+        '--stride',
+        type=float,
+        default=0,
+        help=('the prediction stride equals to '
+              'stride * sample_length, if set as 0, the '
```
innerlee, line 53: Need to define sample_length here

demo/README.md
innerlee: What's the designed range of `STRIDE`? Will `>1` work?

PR #476, opened 2020-12-23
[docs] Update README and readthedocs for ST action detection

dreamerlin:


README.md
```diff
@@ -93,6 +93,12 @@ Supported methods for action localization:
 - [x] [BSN](configs/localization/bsn/README.md)
 - [x] [SSN](configs/localization/ssn/README.md)
 
+Supported methods for spatio-temporal action detection:
+
+- [x] [Fast R-CNN](configs/detection/ava/README.md)
```
kennymckormick, line 98: I think listing SlowOnly and SlowFast is OK, since FastRCNN here is a general pipeline that is used nearly by all ST action detection methods. @innerlee, any suggestion?

innerlee: Writing both together, like `SlowOnly+FastRCNN`?

dreamerlin: so, what's the final decision?

innerlee: I agree with @kennymckormick that FastRCNN should not list here.

We need an policy on whether listing det methods and action methods together. How about
* If the det part and action part are decoupled, namely, the det part could be freely changed to other det methods, then we list both.
* If they are coupled together, being an integrated whole, then we write their own method name only.

innerlee (approved):

PR #492, opened 2020-12-25
[Feature] Support Imgaug for augmentations in the data pipeline.

irvingzhang0512:
Support [imgaug](https://github.com/aleju/imgaug) in dataset pipeline.

+ TODO
    + [x] changelog & docs.
    + [x] tests.

Some of these codes are modified from [mmdet](https://github.com/open-mmlab/mmdetection/blob/482f60fe55c364e50e4fc4b50893a25d8cc261b0/mmdet/datasets/pipelines/transforms.py#L1191). 

Any suggestions about this PR?

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,8 @@
 
 **New Features**
 
+- Support [imgaug](https://imgaug.readthedocs.io/en/latest/index.html) for augmentations in the data pipeline ([#492](https://github.com/open-mmlab/mmaction2/pull/492))
+
 **Improvements**
 
 - Support setting `max_testing_views` for extremely large models to save GPU memory used ([#511](https://github.com/open-mmlab/mmaction2/pull/511))
@@ -29,7 +31,6 @@
 - Support precise BN ([#501](https://github.com/open-mmlab/mmaction2/pull/501/))
 - Support Spatio-Temporal Action Detection (AVA) ([#351](https://github.com/open-mmlab/mmaction2/pull/351))
 - Support to return feature maps in `inference_recognizer` ([#458](https://github.com/open-mmlab/mmaction2/pull/458))
-- Support [imgaug](https://imgaug.readthedocs.io/en/latest/index.html) for augmentations in the data pipeline ([#492](https://github.com/open-mmlab/mmaction2/pull/492))
 
 **Improvements**
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -29,6 +29,7 @@
 - Support precise BN ([#501](https://github.com/open-mmlab/mmaction2/pull/501/))
 - Support Spatio-Temporal Action Detection (AVA) ([#351](https://github.com/open-mmlab/mmaction2/pull/351))
 - Support to return feature maps in `inference_recognizer` ([#458](https://github.com/open-mmlab/mmaction2/pull/458))
+- Support [imgaug](https://imgaug.readthedocs.io/en/latest/index.html) for augmentations in the data pipeline ([#492](https://github.com/open-mmlab/mmaction2/pull/492))
 
 **Improvements**
 
--- a/mmaction/datasets/pipelines/augmentations.py
+++ b/mmaction/datasets/pipelines/augmentations.py
@@ -43,25 +43,24 @@ def _init_lazy_if_proper(results, lazy):
 
 
 @PIPELINES.register_module()
-class Imgaug(object):
+class Imgaug:
     """Imgaug augmentation.
 
     Adds custom transformations from imgaug library.
-    Please, visit `https://imgaug.readthedocs.io/en/latest/index.html`
+    Please visit `https://imgaug.readthedocs.io/en/latest/index.html`
     to get more information. An example of ``transforms`` could be found
-    in `_default_transforms`
+    in `default_transforms`
 
     Required keys are "imgs" and "img_shape"(if "gt_bboxes" is not None),
     added or modified keys are "imgs", "img_shape", "gt_bboxes", "proposals".
 
     It is worth mentioning that `Imgaug` will NOT create custom keys like
     "interpolation", "crop_bbox", "flip_direction", etc. So when using
-    `Imgaug` along with other mmaction2 pipelines, we should pay more attetion
+    `Imgaug` along with other mmaction2 pipelines, we should pay more attention
     to required keys.
 
-    Three steps to use `Imgaug` pipeline:
-    1. Choose our favourite augmenters in imgaug docs and set related args.
-    2. Create initialization parameter `transforms`. There are three ways
+    Two steps to use `Imgaug` pipeline:
+    1. Create initialization parameter `transforms`. There are three ways
         to create `transforms`.
         1) string: only support `default` for now.
             e.g. `transforms='default'`
@@ -73,7 +72,7 @@ class Imgaug(object):
             e.g. `transforms=[dict(type=iaa.Rotate, rotate=(-20, 20))]`
         3) iaa.Augmenter: create an imgaug.Augmenter object.
             e.g. `transforms=iaa.Rotate(rotate=(-20, 20))`
-    3. Add `Imgaug` in dataset pipeline. It is recommended to insert imgaug
+    2. Add `Imgaug` in dataset pipeline. It is recommended to insert imgaug
         pipeline before `FormatShape`. A demo pipeline is listed as follows.
         ```
         pipeline = [
@@ -96,7 +95,7 @@ class Imgaug(object):
             dict(type='Flip', flip_ratio=0.5),
             dict(type='Imgaug', transforms='default'),
             # dict(type='Imgaug', transforms=[
-            #     dict(type=iaa.Rotate, rotate=(-20, 20))
+            #     dict(type='Rotate', rotate=(-20, 20))
             # ]),
             # dict(type='Imgaug',transforms=iaa.Rotate(rotate=(-20, 20))),
             dict(type='FormatShape', input_format='NCHW'),
@@ -117,7 +116,7 @@ class Imgaug(object):
         self.bbs = imgaug.augmentables.bbs
 
         if transforms == 'default':
-            self.transforms = self._default_transforms()
+            self.transforms = self.default_transforms()
         elif isinstance(transforms, list):
             assert all(isinstance(trans, dict) for trans in transforms)
             self.transforms = transforms
@@ -131,7 +130,7 @@ class Imgaug(object):
             self.aug = self.iaa.Sequential(
                 [self.imgaug_builder(t) for t in self.transforms])
 
-    def _default_transforms(self):
+    def default_transforms(self):
         """Default transforms for imgaug."""
 
         return [
@@ -170,8 +169,8 @@ class Imgaug(object):
     def imgaug_builder(self, cfg):
         """Import a module from imgaug.
 
-        It inherits some of :func:`build_from_cfg` logic. Use a dict object to
-        create a iaa.Augmenter object.
+        It follows the logic of :func:`build_from_cfg`. Use a dict object to
+        create an iaa.Augmenter object.
 
         Args:
             cfg (dict): Config dict. It should at least contain the key "type".
--- a/mmaction/datasets/pipelines/augmentations.py
+++ b/mmaction/datasets/pipelines/augmentations.py
@@ -7,14 +7,6 @@ from torch.nn.modules.utils import _pair
 
 from ..registry import PIPELINES
 
-try:
-    import imgaug  # noqa
-    from imgaug import augmenters as iaa
-    from imgaug.augmentables.bbs import BoundingBox, BoundingBoxesOnImage
-except ImportError:
-    imgaug = None
-    iaa = None
-
 
 def _init_lazy_if_proper(results, lazy):
     """Initialize lazy operation properly.
@@ -77,20 +69,22 @@ class Imgaug(object):
     """
 
     def __init__(self, transforms):
-        if imgaug is None or iaa is None:
-            raise RuntimeError('imgaug is not installed')
+        import imgaug
+
+        self.iaa = imgaug.augmenters
+        self.bbs = imgaug.augmentables.bbs
 
         if transforms == 'default':
             self.transforms = self._default_transforms()
         elif isinstance(transforms, list):
             self.transforms = transforms
-        elif isinstance(transforms, iaa.Augmenter):
+        elif isinstance(transforms, self.iaa.Augmenter):
             self.aug = self.transforms = transforms
         else:
             raise ValueError("transforms must be 'default' or a list of dicts")
 
-        if not isinstance(transforms, iaa.Augmenter):
-            self.aug = iaa.Sequential(
+        if not isinstance(transforms, self.iaa.Augmenter):
+            self.aug = self.iaa.Sequential(
                 [self.imgaug_builder(t) for t in self.transforms])
 
     def _default_transforms(self):
@@ -148,9 +142,9 @@ class Imgaug(object):
 
         obj_type = args.pop('type')
         if mmcv.is_str(obj_type):
-            obj_cls = getattr(iaa, obj_type)
+            obj_cls = getattr(self.iaa, obj_type)
         elif inspect.isclass(obj_type):
-            assert obj_type == getattr(iaa, obj_type.__name__)
+            assert obj_type == getattr(self.iaa, obj_type.__name__)
             obj_cls = obj_type
         else:
             raise TypeError(
@@ -177,10 +171,11 @@ class Imgaug(object):
 
         if 'gt_bboxes' in results:
             bbox_list = [
-                BoundingBox(x1=bbox[0], y1=bbox[1], x2=bbox[2], y2=bbox[3])
+                self.bbs.BoundingBox(
+                    x1=bbox[0], y1=bbox[1], x2=bbox[2], y2=bbox[3])
                 for bbox in results['gt_bboxes']
             ]
-            bboxes = BoundingBoxesOnImage(
+            bboxes = self.bbs.BoundingBoxesOnImage(
                 bbox_list, shape=results['img_shape'])
             bbox_aug = cur_aug.augment_bounding_boxes([bboxes])[0]
             results['gt_bboxes'] = [[
@@ -191,11 +186,11 @@ class Imgaug(object):
             ] for bbox in bbox_aug.items]
             if 'proposals' in results:
                 bbox_list = [
-                    BoundingBox(
+                    self.bbs.BoundingBox(
                         x1=bbox[0], y1=bbox[1], x2=bbox[2], y2=bbox[3])
                     for bbox in results['proposals']
                 ]
-                bboxes = BoundingBoxesOnImage(
+                bboxes = self.bbs.BoundingBoxesOnImage(
                     bbox_list, shape=results['img_shape'])
                 bbox_aug = cur_aug.augment_bounding_boxes([bboxes])[0]
                 results['proposals'] = [[
--- a/mmaction/datasets/pipelines/augmentations.py
+++ b/mmaction/datasets/pipelines/augmentations.py
@@ -98,7 +98,6 @@ class Imgaug:
             # dict(type='Imgaug', transforms=[
             #     dict(type='Rotate', rotate=(-20, 20))
             # ]),
-            # dict(type='Imgaug',transforms=iaa.Rotate(rotate=(-20, 20))),
             dict(type='FormatShape', input_format='NCHW'),
             dict(type='Collect', keys=['imgs', 'label'], meta_keys=[]),
             dict(type='ToTensor', keys=['imgs', 'label'])
--- a/mmaction/datasets/pipelines/augmentations.py
+++ b/mmaction/datasets/pipelines/augmentations.py
@@ -48,8 +48,13 @@ class Imgaug:
 
     Adds custom transformations from imgaug library.
     Please visit `https://imgaug.readthedocs.io/en/latest/index.html`
-    to get more information. An example of ``transforms`` could be found
-    in `default_transforms`
+    to get more information. Two demo configs could be found in tsn and i3d
+    config folder.
+
+    It's better to use uint8 images as inputs since imgaug works best with
+    numpy dtype uint8 and isn't well tested with other dtypes. It should be
+    noted that not all of the augmenters have the same input and output dtype,
+    which may cause unexpected results.
 
     Required keys are "imgs", "img_shape"(if "gt_bboxes" is not None) and
     "modality", added or modified keys are "imgs", "img_shape", "gt_bboxes"
@@ -74,7 +79,7 @@ class Imgaug:
         3) iaa.Augmenter: create an imgaug.Augmenter object.
             e.g. `transforms=iaa.Rotate(rotate=(-20, 20))`
     2. Add `Imgaug` in dataset pipeline. It is recommended to insert imgaug
-        pipeline before `FormatShape`. A demo pipeline is listed as follows.
+        pipeline before `Normalize`. A demo pipeline is listed as follows.
         ```
         pipeline = [
             dict(
@@ -98,6 +103,7 @@ class Imgaug:
             # dict(type='Imgaug', transforms=[
             #     dict(type='Rotate', rotate=(-20, 20))
             # ]),
+            dict(type='Normalize', **img_norm_cfg),
             dict(type='FormatShape', input_format='NCHW'),
             dict(type='Collect', keys=['imgs', 'label'], meta_keys=[]),
             dict(type='ToTensor', keys=['imgs', 'label'])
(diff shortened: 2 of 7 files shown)


innerlee:
This library looks interesting.

Do you know the typical use cases? Any mixed usage of imgaug and normal augmentation pipelines?

irvingzhang0512:
Imgaug (and most of the other data augmentation libs) is desined for single image augmentation(eg. blur/rotate/crop/pad/resize...), along with all kinds of labels(detection, segmentation, keypoints). So when dealing ava dataset, no need to take care of bbox separately

+ The usage is quite simple:
    + step1: find our favourite augmenters in [imgaug docs](https://imgaug.readthedocs.io/en/latest/source/overview_of_augmenters.html).
    + step2: create `transforms`.
    + step3: add imgaug pipeline before `FormatShape`.

```python
pipeline = [
    dict(
        type='SampleFrames',
        clip_len=1,
        frame_interval=1,
        num_clips=16,
    ),
    dict(type='RawFrameDecode'),
    dict(type='Resize', scale=(-1, 256)),
    dict(
        type='MultiScaleCrop',
        input_size=224,
        scales=(1, 0.875, 0.75, 0.66),
        random_crop=False,
        max_wh_scale_gap=1,
        num_fixed_crops=13),
    dict(type='Resize', scale=(224, 224), keep_ratio=False),
    dict(type='Flip', flip_ratio=0.5),
    dict(type='Imgaug', transforms='default'),
    # dict(type='Imgaug', transforms=[dict(type='Rotate', rotate=(-30, 30)]),
    dict(type='FormatShape', input_format='NCHW'),
    dict(type='Collect', keys=['imgs', 'label'], meta_keys=[]),
    dict(type='ToTensor', keys=['imgs', 'label'])
]
```

However, imgaug pipeline only generates/modifies 4 keys `imgs/img_shape/gt_bboxes/proposals`. Normal pipelines will also generate keys like crop_bbox, flip_direction, etc...

I was planing to create a config file,  using imgaug to replace all normal augmentation pipelines. But imgaug doesn't support all mmaction2 augmentation pipelines, especially the crop/resize. 

After this pr is merged, I'll also create a pr to support [albumentations](https://github.com/albumentations-team/albumentations), which is also a third party data augmentation library and more popular nowadays.

irvingzhang0512:
FYI, [mmdet pr about albumentations](https://github.com/open-mmlab/mmdetection/pull/1354)

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +50,166 @@ def _init_lazy_if_proper(results, lazy):
...
+    Required keys are "imgs" and "img_shape"(if "gt_bboxes" is not None),
+    added or modified keys are "imgs", "img_shape", "gt_bboxes", "proposals".
+
+    It is worth mentioning that `Imgaug` will NOT create custom keys like
+    "interpolation", "crop_bbox", "flip_direction", etc. So when using
+    `Imgaug` along with other mmaction2 pipelines, we should pay more attetion
+    to required keys of all pipelines.
+
+    Args:
+        transforms (str or list[dict] or iaa.Augmenter): Three different ways
```
SuX97, line 71:
```
transforms (str | list[dict] | :obj:`iaa.Augmenter`)
```

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +50,166 @@ def _init_lazy_if_proper(results, lazy):
...
+    """
+
+    def __init__(self, transforms):
+        if imgaug is None or iaa is None:
+            raise RuntimeError('imgaug is not installed')
+
+        if transforms == 'default':
+            self.transforms = self._default_transforms()
+        elif isinstance(transforms, list):
+            self.transforms = transforms
```
SuX97, line 86: Maybe assert here that it is a list of dict?

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +50,166 @@ def _init_lazy_if_proper(results, lazy):
...
+        """
+        import inspect
+
+        assert isinstance(cfg, dict) and 'type' in cfg
+        args = cfg.copy()
+
+        obj_type = args.pop('type')
+        if mmcv.is_str(obj_type):
+            obj_cls = getattr(iaa, obj_type)
+        elif inspect.isclass(obj_type):
```
SuX97, line 152: Is it possible for users to use this 'instantiation by cls' by simply modify the config file? Maybe add a new tutorial for this.

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +50,166 @@ def _init_lazy_if_proper(results, lazy):
...
+            ]
+
+        return obj_cls(**args)
+
+    def __repr__(self):
+        repr_str = self.__class__.__name__ + f'(transforms={self.transforms})'
+        return repr_str
+
+    def __call__(self, results):
+        cur_aug = self.aug.to_deterministic()
```
SuX97, line 171: Just curious: does it means perform the same aug (such as RandomCrop) for each imgs?

tests/test_data/test_augmentations.py
```diff
@@ -1230,3 +1230,71 @@ def test_box_flip(self):
...
+
+        with pytest.raises(AttributeError):
+            # `type` must be available in imgaug
+            Imgaug(transforms=[dict(type='BlaBla')])
+
+        from imgaug import augmenters as iaa
+
+        # check flip (both images and bboxes)
+        target_keys = ['imgs', 'gt_bboxes', 'proposals', 'img_shape']
+        imgs = list(np.random.rand(1, 305, 200, 3))
```
SuX97, line 1256: Use smaller images for faster CI testing speed.

mmaction/datasets/pipelines/augmentations.py
irvingzhang0512: I'll take a look at imgaug source code to find out.

mmaction/datasets/pipelines/augmentations.py
irvingzhang0512: yes

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +50,166 @@ def _init_lazy_if_proper(results, lazy):
...
+
+        if 'children' in args:
+            args['children'] = [
+                self.imgaug_builder(child) for child in args['children']
+            ]
+
+        return obj_cls(**args)
+
+    def __repr__(self):
+        repr_str = self.__class__.__name__ + f'(transforms={self.transforms})'
```
SuX97, line 167: The action of this will be different when using different kinds of instantiations. I recommend using a unified form i.e. build from config dict.

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +50,166 @@ def _init_lazy_if_proper(results, lazy):
...
+            to create imgaug augmenter.
+            1) string: only support one string `default`
+            2) list[dict]: create a list of Augmenters by a list of dicts.
+                Each dict contains a key named `type`.
+            3) iaa.Augmenter: Directly use imgaug class.
+    """
+
+    def __init__(self, transforms):
+        if imgaug is None or iaa is None:
+            raise RuntimeError('imgaug is not installed')
```
SuX97, line 81: It is not tested, and maybe never be tested as in #L81 it will throw 'referenced before assigned error'. Maybe just leave it be as you protect the import at the beginning.

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +50,166 @@ def _init_lazy_if_proper(results, lazy):
...
+        elif isinstance(transforms, list):
+            self.transforms = transforms
+        elif isinstance(transforms, iaa.Augmenter):
+            self.aug = self.transforms = transforms
+        else:
+            raise ValueError("transforms must be 'default' or a list of dicts")
+
+        if not isinstance(transforms, iaa.Augmenter):
+            self.aug = iaa.Sequential(
+                [self.imgaug_builder(t) for t in self.transforms])
```
SuX97, line 94: Does it mean the input transforms can also be `list[iaa.Augmenter]`?

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +50,166 @@ def _init_lazy_if_proper(results, lazy):
...
+        img_h, img_w, _ = results['imgs'][0].shape
+
+        if 'gt_bboxes' in results:
+            bbox_list = [
+                BoundingBox(x1=bbox[0], y1=bbox[1], x2=bbox[2], y2=bbox[3])
+                for bbox in results['gt_bboxes']
+            ]
+            bboxes = BoundingBoxesOnImage(
+                bbox_list, shape=results['img_shape'])
+            bbox_aug = cur_aug.augment_bounding_boxes([bboxes])[0]
```
SuX97, line 185: What are the other indexes other than 0? Maybe use
```
bbox_aug, *_ = cur_aug.augment_bounding_boxes([bboxes])
```
for readability as others are not familiar with iaa.

mmaction/datasets/pipelines/augmentations.py
SuX97:
```
pipeline = [
    dict(
        type='SampleFrames',
        clip_len=1,
        frame_interval=1,
        num_clips=16,
    ),
    dict(type='RawFrameDecode'),
    dict(type='Resize', scale=(-1, 256)),
    dict(
        type='MultiScaleCrop',
        input_size=224,
        scales=(1, 0.875, 0.75, 0.66),
        random_crop=False,
        max_wh_scale_gap=1,
        num_fixed_crops=13),
    dict(type='Resize', scale=(224, 224), keep_ratio=False),
    dict(type='Flip', flip_ratio=0.5),
    dict(type='Imgaug', transforms='default'),
    # dict(type='Imgaug', transforms=[dict(type='Rotate', rotate=(-30, 30))]),
    dict(type='FormatShape', input_format='NCHW'),
    dict(type='Collect', keys=['imgs', 'label'], meta_keys=[]),
    dict(type='ToTensor', keys=['imgs', 'label'])
]
```
Maybe show a way to build it from config such as this. Seems that you need to import iaa in the config file.

SuX97:
That's interesting. If you have trained some models that benefit from this augmentation, could you please share some with us?

BTW, there are many other augmentation packages such as https://github.com/hassony2/torch_videovision, https://github.com/torchvideo/torchvideo, https://github.com/DemisEom/SpecAugment that might be helpful.

Especially this one https://github.com/piergiaj/pytorch-i3d/blob/master/videotransforms.py, which applies augmentation by volume slicing rather than for looping. @innerlee Will it be faster?

innerlee:
no

SuX97:
> no
@lizz , I am not sure about that. I have done a toy test:
```
import numpy as np

def tiktok(func):
	import time
	def wrapper(*args, **kwargs):
		tik = time.time()
		ret = func(*args, **kwargs)
		tok = time.time()
		print(f'{func.__name__} takes {tok - tik}s.')
		return ret
	return wrapper

@tiktok
def slice(imgs):
	for i in range(1000):
		_ = imgs[:, 5:219, 5:219, :]

@tiktok
def for_loop(imgs):
	for i in range(1000):
		out = [img[5:219, 5:219, :] for img in imgs]
		#out = np.array(out)

imgs = np.random.randn(32, 256, 256, 3)
slice(imgs)
for_loop(imgs)
```
This gives a result of:
slice takes 0.0004169940948486328s.
for_loop takes 0.018939733505249023s.

Since we do the for-loop slicing and convert it into numpy array afterward, when taking ```out = np.array(out)``` into account, it yields:
slice takes 0.0003859996795654297s.
for_loop takes 8.275573968887329s.

mmaction/datasets/pipelines/augmentations.py
irvingzhang0512: No, `imgaug_builder` passes a  `dict` object as input, and returns a `iaa.Augmenter` object. I'll update `imgaug_builder` docs about this.

And I don't think it's nesseary to suport `list[iaa.Augmenter]`. If we do have a `list[iaa.Augmenter]`, we can simply create a `iaa.Sequential(list_of_iaa_augmenters)` as transforms.

irvingzhang0512:
I haven't tried these augmenters to train kinetics-400. When training our own datasets, blur/rotate helps.

My original intention is to try everything. For now, speed is not my priority. So I prefer 3rd party libraries with more augmenters(imgaug and albumentations). If we can all benefit from certain augmenters, we can implements those in the future.

BTW, `albumentations` doesn't use loop. I'll take a close look at related source codes later.

mmaction/datasets/pipelines/augmentations.py
irvingzhang0512: `aug.augment_bounding_boxes` and `aug.augment_images` are designed for augmenting a batch of images. Each image has its corresponding bboxes and each image has different augmenter args. 

A typical use case is shown as follows. A/Abb, B/Bbb, C/Cbb have different rotate degree.

```python
Abb = Bbb = Cbb = BoundingBoxesOnImage([
    BoundingBox(1, 1, 9, 9)], (10, 10))
seq = iaa.Rotate()
seq_det = seq.to_deterministic()
imgs_aug = seq_det.augment_images([A, B, C])
bbs_aug = seq_det.augment_bounding_boxes([Abb, Bbb, Cbb])
```

We use `augment_image` instead of `augment_images` to use same aug for all images.

mmaction/datasets/pipelines/augmentations.py
irvingzhang0512: I misunderstood you last night... Add some tutorials in class comments.

If `type` is iaa.Augmenter subclass, we need to import iaa first. If we use string `type`, there is no need to import iaa.

docs/changelog.md
```diff
@@ -8,6 +8,7 @@
 
 - Support Spatio-Temporal Action Detection(AVA) ([#351](https://github.com/open-mmlab/mmaction2/pull/351))
 - Support to return feature maps in `inference_recognizer` ([#458](https://github.com/open-mmlab/mmaction2/pull/458))
+- Support [imgaug](https://imgaug.readthedocs.io/en/latest/index.html) for augmentations in the data pipeline ([#492](https://github.com/open-mmlab/mmaction2/pull/492))
```
SuX97, line 11: Please add to v0.11 changelog.

SuX97:
you may need to pull the new master and rebase to it.

irvingzhang0512:
kindly ping @dreamerlin @SuX97

SuX97 (approved):

irvingzhang0512:
@innerlee any suggestions for this pr?

innerlee:
Quite busy this week, will take a look next Wed. Thanks for the patience

irvingzhang0512:
kindly ping @innerlee

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,210 @@ def _init_lazy_if_proper(results, lazy):
         assert 'lazy' not in results, 'Use Fuse after lazy operations'
 
 
+@PIPELINES.register_module()
+class Imgaug(object):
+    """Imgaug augmentation.
+
+    Adds custom transformations from imgaug library.
+    Please, visit `https://imgaug.readthedocs.io/en/latest/index.html`
```
innerlee, line 50: `Please, visit` -> `Please visit`

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,210 @@ def _init_lazy_if_proper(results, lazy):
...
+    to get more information. An example of ``transforms`` could be found
+    in `_default_transforms`
```
innerlee, line 52: If `_default_transforms` is intened to be visible from outside, then the leading `_` should be removed from the name

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,210 @@ def _init_lazy_if_proper(results, lazy):
...
+
+    Required keys are "imgs" and "img_shape"(if "gt_bboxes" is not None),
+    added or modified keys are "imgs", "img_shape", "gt_bboxes", "proposals".
+
+    It is worth mentioning that `Imgaug` will NOT create custom keys like
+    "interpolation", "crop_bbox", "flip_direction", etc. So when using
+    `Imgaug` along with other mmaction2 pipelines, we should pay more attetion
```
innerlee, line 59: typo attetion

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,210 @@ def _init_lazy_if_proper(results, lazy):
...
+    to required keys.
+
+    Three steps to use `Imgaug` pipeline:
+    1. Choose our favourite augmenters in imgaug docs and set related args.
```
innerlee, line 63: Step 1 can be removed?

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,210 @@ def _init_lazy_if_proper(results, lazy):
...
+                input_size=224,
+                scales=(1, 0.875, 0.75, 0.66),
+                random_crop=False,
+                max_wh_scale_gap=1,
+                num_fixed_crops=13),
+            dict(type='Resize', scale=(224, 224), keep_ratio=False),
+            dict(type='Flip', flip_ratio=0.5),
+            dict(type='Imgaug', transforms='default'),
+            # dict(type='Imgaug', transforms=[
+            #     dict(type=iaa.Rotate, rotate=(-20, 20))
```
innerlee, line 99: Use string version ` "Rotate" ` as example instead of `iaa.Rotate` since we do not want import libs in config files.

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,210 @@ def _init_lazy_if_proper(results, lazy):
         assert 'lazy' not in results, 'Use Fuse after lazy operations'
 
 
+@PIPELINES.register_module()
+class Imgaug(object):
```
innerlee, line 46: `class Imgaug:`

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,210 @@ def _init_lazy_if_proper(results, lazy):
...
+                        loc=0,
+                        scale=(0.0, 0.05 * 255),
+                        per_channel=0.5),
+                ]),
+        ]
+
+    def imgaug_builder(self, cfg):
+        """Import a module from imgaug.
+
+        It inherits some of :func:`build_from_cfg` logic. Use a dict object to
```
innerlee, line 173: `inherits some of` -> `resembles/borrows/follows the`

mmaction/datasets/pipelines/augmentations.py
innerlee: Follow up question: does different videos have different aug? and how?

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,210 @@ def _init_lazy_if_proper(results, lazy):
...
+
+    def __repr__(self):
+        repr_str = self.__class__.__name__ + f'(transforms={self.aug})'
+        return repr_str
+
+    def __call__(self, results):
+        cur_aug = self.aug.to_deterministic()
+
+        results['imgs'] = [
+            cur_aug.augment_image(frame) for frame in results['imgs']
```
innerlee, line 209: It seems that this one does not support optical flow. So an assertion would be great to prevent user accidently got wrong results for flows.

mmaction/datasets/pipelines/augmentations.py
irvingzhang0512: Yes, different videos have diffent aug.

1. `self.aug` doen't have deterministic augmenters. So if we use `self.aug` to augment images, each image has different augs.
2. In order to augment images from a video with the same augmenters, we use `cur_aug = self.aug.to_deterministic()`.
3. Every time we call `self.aug.to_deterministic()`, we get different augs.

```python
# aug1 & aug2 are different.
aug1 = self.aug.to_deterministic()
aug2 = self.aug.to_deterministic()
```

mmaction/datasets/pipelines/augmentations.py
irvingzhang0512: I'm not familar with optical flow. I thought optimal flows are normal grayscale images. Why optical flow was not supported here?

innerlee: Optical flow saves offset at each pixel. Say if you flip horizontally, then the offset need to be negated other than be flipped.

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,209 @@ def _init_lazy_if_proper(results, lazy):
         assert 'lazy' not in results, 'Use Fuse after lazy operations'
 
 
+@PIPELINES.register_module()
+class Imgaug:
+    """Imgaug augmentation.
```
innerlee, line 47: What's the input & output type of this op? uint8 np array? float32? float tensor?

irvingzhang0512: Different augmenters support different dtyps. Overall, `uint8` supports all augmenters.

Details could be found [here](https://imgaug.readthedocs.io/en/latest/source/dtype_support.html)

innerlee: I care about the type change. If we input an uint8, will the output be guaranted to be uint8? Or the output will be float64 regardless of input?

innerlee: To be more specific, we do not want the output dtype to change due to random switches in the processing pipeline.

irvingzhang0512: Based on my experience, dype won't change. But I cannot find any docs about this.

irvingzhang0512: https://github.com/aleju/imgaug/issues/110#issuecomment-375043766
The above comment(2018.5) shows that most of the augmenters will preserve input dtype. And we should use uint8 images as input.

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,212 @@ def _init_lazy_if_proper(results, lazy):
...
+                random_crop=False,
+                max_wh_scale_gap=1,
+                num_fixed_crops=13),
+            dict(type='Resize', scale=(224, 224), keep_ratio=False),
+            dict(type='Flip', flip_ratio=0.5),
+            dict(type='Imgaug', transforms='default'),
+            # dict(type='Imgaug', transforms=[
+            #     dict(type='Rotate', rotate=(-20, 20))
+            # ]),
+            # dict(type='Imgaug',transforms=iaa.Rotate(rotate=(-20, 20))),
```
innerlee, line 101: remove this line since it contains `iaa.blablah`. We do not want to indicate that configs can import pkgs.

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +42,211 @@ def _init_lazy_if_proper(results, lazy):
...
+                max_wh_scale_gap=1,
+                num_fixed_crops=13),
+            dict(type='Resize', scale=(224, 224), keep_ratio=False),
+            dict(type='Flip', flip_ratio=0.5),
+            dict(type='Imgaug', transforms='default'),
+            # dict(type='Imgaug', transforms=[
+            #     dict(type='Rotate', rotate=(-20, 20))
+            # ]),
+            dict(type='FormatShape', input_format='NCHW'),
+            dict(type='Collect', keys=['imgs', 'label'], meta_keys=[]),
```
innerlee, line 102: where is Normalize  in this example?

innerlee:
It will be great to add an example (runnable) config in tsn/ or i3d/ folder

mmaction/datasets/pipelines/augmentations.py
innerlee: Okay, this info can be added to docstring

irvingzhang0512:
Found a serious bug: training failed when using imgaug and multi-gpu. Works fine when training with single-gpu(i used to do).  Will try to fix this in two weeks.


```
  File "./tools/train.py", line 178, in <module>
    main()
  File "./tools/train.py", line 174, in main
    meta=meta)
  File "/ssd/zhangyiyang/mmaction2/mmaction/apis/train.py", line 156, in train_model
    runner.run(data_loaders, cfg.workflow, cfg.total_epochs, **runner_kwargs)
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/site-packages/mmcv/runner/epoch_based_runner.py", line 125, in run
    epoch_runner(data_loaders[i], **kwargs)
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/site-packages/mmcv/runner/epoch_based_runner.py", line 47, in train
    for i, data_batch in enumerate(self.data_loader):
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 291, in __iter__
    return _MultiProcessingDataLoaderIter(self)
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 737, in __init__
    w.start()
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/multiprocessing/process.py", line 112, in start
    self._popen = self._Popen(self)
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/multiprocessing/context.py", line 223, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/multiprocessing/popen_fork.py", line 20, in __init__
    self._launch(process_obj)
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/home/user/.conda/envs/zyy_pytorch1.6/lib/python3.7/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
TypeError: can't pickle module objects
```

innerlee:
try putting import in call function instead of init

innerlee (approved):

innerlee:
@congee524 Please train a checkpoint for these two configs, and upload them in a separate pr, thx!

PR #495, opened 2020-12-28
[Fix] Fix CI bug due to mmcv index

kennymckormick:


innerlee:
`-q` seems unrelated.

kennymckormick:
> `-q` seems unrelated.

just to check log, seems if -q added, nothing will be printed

innerlee:
Is this caused by mmdet? The CI was successful before

kennymckormick:
> Is this caused by mmdet? The CI was successful before

Seems the mmcv with wrong version is installed in the cuda CI, see 2 screenshots below:
image
image

Try to find out why now.

innerlee:
Try to modify the cmd format first

image

mmaction/models/backbones/resnet3d.py
```diff
@@ -999,5 +1000,5 @@ def train(self, mode=True):
                     m.eval()
 
 
-if 'mmdet' in dir():
+if mmdet_imported:
     MMDET_SHARED_HEADS.register_module()(ResNet3dLayer)
```
hellock, line 1004: This line is too weird. The following is more straight forward.

```python
MMDET_SHARED_HEADS.register_module(ResNet3dLayer)
```

kennymckormick: However, registry.register_module() return a function, 
```python
@MMDET_SHARED_HEADS.register_module()
class ResNet3dLayer():
    pass
```
is equivalent to: 
```python
MMDET_SHARED_HEADS.register_module()(ResNet3dLayer)
```

innerlee:
CI has been fixed in master, so you can remove changes in build.yaml

kennymckormick:
> CI has been fixed in master, so you can remove changes in build.yaml

Maybe you can still merge my modification. Since there is no built wheel for torch1.5.1 (which means you still can not pass the GPU CI), I modify build.yml to run CI under torch1.5.0 + torchvision 0.6.0.

kennymckormick:
> Is this caused by mmdet? The CI was successful before

Think so. Bu

> CI has been fixed in master, so you can remove changes in build.yaml

Seems although https://download.openmmlab.com/mmcv/dist/cu101/torch1.5.1/index.html is an empty page, you can still install mmcv-full and pass the GPU CI. I'm not sure how does that happen, would you like to share the explanation with me?

innerlee:
🤷 have no idea.
Ask @ycxioooong

ycxioooong:
@kennymckormick There is no built wheel for torch1.5.1 from a long time ago. We may have a discussion on this issue offline.

dreamerlin:
How is it going ? @kennymckormick

kennymckormick:
> How is it going ? @kennymckormick

I think OK to merge.

innerlee (approved):

PR #501, opened 2020-12-29
Add Precise BN Hook

SuX97:


mmaction/utils/precise_bn.py
```diff
@@ -0,0 +1,156 @@
...
+        model (nn.Module): The model whose bn stats will be recomputed.
+        data_loader (iterator): The DataLoader iterator.
+        num_iters (int): number of iterations to compute the stats.
+        logger (:obj:`logging.Logger` | None): Logger for logging.
+            Default: None.
+    """
+
+    model.train()
+
+    assert len(data_loader) >= num_iters, (
```
kennymckormick, line 60: Since you set the attribute `interval` for PreciseBNHook, can we change the condition to 
`assert interval * len(data_loader) >= num_iters`?

SuX97: I'm not sure. `interval` means updating bn stats every `interval` epochs. While in this assertion, `num_iters` and `len(data_loader)` ensures for EACH epoch the steps is valid.

Do you mean by setting a total number of iters for precise bn calculation? Like if I want to do it every 5 epochs for 50 epochs. The total iters of precise bn should be (50 // 5) * len(data_loader).

SuX97: I think this way is more straight-forward. And it is consistent with the original implementation of fvcore.

kennymckormick: Actually, I mean if one wants to do update_bn_stats every 5 epochs, and each epoch has 60 batches. Then num_iters can be at most 300 instead of 60. Maybe useful for some small datasets.

SuX97: 300 iterations will have the same effect as 60 iterations, as other parameters in the models are fixed except bn stats. 

It follows the original setting https://github.com/facebookresearch/fvcore/blob/1c6e25b1dd4d9de8af7168faabbcedb83e629c8e/fvcore/nn/precise_bn.py#L91

kennymckormick (approved):

SuX97:
w/ precise bn: slowfast 4x16 acc1/5: 74.75/91.59

w/o precise bn: slowfast 4x16 acc1/5: 74.75/91.73

No significant improvement. Similar to the effect on TSN, SlowOnly.

Guess the distribution of EMA calculated stats is similar to that of precise bn. Still, it might be helpful for users who have a severe domain gap between train and val. @innerlee

mmaction/utils/precise_bn.py
```diff
@@ -0,0 +1,156 @@
+# Modified from https://github.com/facebookresearch/fvcore/blob/master/fvcore/nn/precise_bn.py  # noqa
```
innerlee, line 1: What the original license?

SuX97: Apache License 2.0

innerlee: Need to add this piece of info

mmaction/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 from .version import __version__
 
-mmcv_minimum_version = '1.1.1'
+mmcv_minimum_version = '1.2.4'
```
innerlee, line 6: does this feature relies on v1.2.4?

SuX97: Nah, just to get passed with the CI.

PR #504, opened 2020-12-30
add lazy ops docs

### LINKED ISSUE #418: Need docs about lazy ops
**Describe the feature**
Need to write features about the lazy ops on docs to indicate this efficient things. 

**Motivation**
Make users know about it clearly

**Addition**
Try to add lazy feature applied to other transformation.

congee524:
fix #418

### DIFF
--- a/docs/tutorials/4_data_pipeline.md
+++ b/docs/tutorials/4_data_pipeline.md
@@ -87,6 +87,39 @@ test_pipeline = [
 ]
 ```
 
+We have supported some lazy operators and encourage users to apply them.
+Lazy ops record how the data should be processed, but it will postpone the processing on the raw data until the raw data forward `Fuse` stage.
+Specifically, lazy ops avoid frequent reading and modification operation on the raw data, but process the raw data once in the final Fuse stage, thus accelerating data preprocessing greatly.
+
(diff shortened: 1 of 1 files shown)


docs/tutorials/4_data_pipeline.md
```diff
@@ -87,6 +87,39 @@ test_pipeline = [
 ]
 ```
 
+We have supported some lazy operators and encourage users to apply them.
+Lazy ops record how the data should be processed, but it will postpone the processing on the raw data until the raw data forward `Fuse` stage.
+Specifically, lazy ops avoid frequent reading and modification operation on the raw data, but process the raw data once in the final Fuse stage, thus accelerating data preprocessing greatly.
```
innerlee, line 92: Any numbers for `greatly`?

congee524: The effect on 320p K400 is obvious (but it could also be the memcache's efffect). I'll remove `greatly` frist. A comparison result may be posted in the future.

congee524: @dreamerlin Do you remember the experiment？

innerlee (approved):

PR #505, opened 2020-12-30
boost bmn speed for parrots

jiaoml1996:
booast bmn speed for parrots

### DIFF
--- a/mmaction/models/localizers/bmn.py
+++ b/mmaction/models/localizers/bmn.py
@@ -134,10 +134,7 @@ class BMN(BaseLocalizer):
         self.anchors_tmins, self.anchors_tmaxs = self._temporal_anchors(
             -0.5, 1.5)
         self.match_map = self._match_map()
-        if torch.__version__ == 'parrots':
-            self.bm_mask = self._get_bm_mask().cuda()
-        else:
-            self.bm_mask = self._get_bm_mask()
+        self.bm_mask = self._get_bm_mask().cuda()
 
     def _match_map(self):
         """Generate match map."""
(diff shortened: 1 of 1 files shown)


mmaction/models/localizers/bmn.py
```diff
@@ -134,7 +134,7 @@ def __init__(self,
         self.anchors_tmins, self.anchors_tmaxs = self._temporal_anchors(
             -0.5, 1.5)
         self.match_map = self._match_map()
-        self.bm_mask = self._get_bm_mask()
+        self.bm_mask = self._get_bm_mask().cuda()
```
innerlee, line 137: Is this a bug fix?

innerlee: Okay, could you provide insight on why this version is faster?

jiaoml1996: Because the tensor.cuda() hostToDevice of parrots is very slow, we calculate asynchronously.

mmaction/models/localizers/bmn.py
```diff
@@ -135,6 +135,8 @@ def __init__(self,
             -0.5, 1.5)
         self.match_map = self._match_map()
         self.bm_mask = self._get_bm_mask()
+        if torch.cuda.is_aviaiable():
+            self.bm_mask = self.bm_mask.cuda()
```
innerlee, line 139: Will this work when using multiple gpus? Can it go to the correct device?

jiaoml1996: It can work normally with multiple gpus.

kennymckormick:
The problem is, this repo seems to support DataParallel (but used nowhere, actually). I think `.cuda()` might not support DataParallel with every gpu. Can we disable DataParallel? So that `.cuda()` will be safe in every case. @innerlee @jiaoml1996

innerlee:
DataParallel is more familiar to some users since it is easier to setup and use. I wouldn't suggest disableing it.

Can this fix be smarter to figure out which device to go?

kennymckormick:
Unfortunately, since the module is replicated on each gpu in each forward, for DataParallel, you have to call bm_mask.to(raw_feature.device) in every single forward.

dreamerlin (approved):

kennymckormick:
@dreamerlin I think this PR can not be merged now since:
1. It is not compatible with DataParallel with multiple GPUs
2. It is not compatible with CPU mode: (`torch.cuda.is_available()` only indicates that you can run gpu mode, not you are running gpu mode)

innerlee:
@jiaoml1996 Do you have other implementations that can figure out the goto device correctly? As @kennymckormick pointed out, the current version is not friendly to cpu/gpu switching.

dreamerlin:
kindly ping @jiaoml1996

dreamerlin:
Close this PR and will open again when there are any updates

PR #509, opened 2021-01-04
Add class weight

kennymckormick:
1. Add class_weight for CrossEntropyLoss and BCELossWithLogits
2. switch mmcv to 1.2.4 in CI


mmaction/models/losses/cross_entropy_loss.py
```diff
@@ -20,13 +36,30 @@ def _forward(self, cls_score, label, **kwargs):
         Returns:
             torch.Tensor: The returned CrossEntropy loss.
         """
+        if self.class_weight is not None:
+            kwargs['weight'] = self.class_weight.to(cls_score.device)
```
innerlee, line 40: assert that `weight` do not present in kwargs before this line?

mmaction/models/losses/cross_entropy_loss.py
```diff
@@ -40,6 +73,8 @@ def _forward(self, cls_score, label, **kwargs):
         Returns:
             torch.Tensor: The returned bce loss with logits.
         """
+        if self.class_weight is not None:
+            kwargs['weight'] = self.class_weight.to(cls_score.device)
```
innerlee, line 77: same here

mmaction/models/losses/cross_entropy_loss.py
kennymckormick: It does not present elsewhere. But adding an assert line here is a good idea.

innerlee (approved):

PR #511, opened 2021-01-04
[Feature] Test batch

kennymckormick:
Add feature `test_batch` for extremely large model testing.
Background: 
For extremely large models, you can not even forward one sample at a time (which includes 10x3=30 views). By setting test_batch in test_cfg, you can forward `test_batch` views at a time.


mmaction/models/recognizers/recognizer3d.py
```diff
@@ -29,13 +31,28 @@ def _do_test(self, imgs):
...
         imgs = imgs.reshape((-1, ) + imgs.shape[2:])
 
-        x = self.extract_feat(imgs)
-        if hasattr(self, 'neck'):
-            x, _ = self.neck(x)
+        if self.test_batch is not None:
+            tot = imgs.shape[0]
+            assert num_segs == tot, ('test_batch is only compatible with '
+                                     'batch_size == 1')
+            ptr = 0
```
dreamerlin, line 38: `ptr` and `total` are not good names. How about `batch_point` and `total_batches` ?

dreamerlin:
We may need some explanation for this feature in docs.

mmaction/models/recognizers/recognizer3d.py
kennymckormick: I will change it to view_ptr and total_views, I think `ptr` is widely used to represent a pointer.

innerlee:
As discussed, we change the name to `max_testing_views`

innerlee (approved):

PR #516, opened 2021-01-05
Remove links of MiT and MMiT

kennymckormick:


tools/data/mit/preprocess_data.sh
```diff
@@ -9,7 +9,7 @@ fi
 
 cd ${DATA_DIR}
 
-wget http://data.csail.mit.edu/soundnet/actions3/split1/Moments_in_Time_Raw.zip
+# Download the Moments_in_Time_Raw.zip here manually
```
SuX97, line 12: Maybe leave a runnable api for users who assume runnable:

$DOWNLOAD_URL=$1
wget  $DOWNLOAD_URL Moments_in_Time_Raw.zip

And the zip name may change so manually assign it `Moments_in_Time_Raw.zip`

tools/data/mit/README.md
```diff
@@ -20,11 +20,7 @@ Before we start, please make sure that the directory is located at `$MMACTION2/t
 
 ## Step 1. Prepare Annotations and Videos
 
-First of all, you can run the following script to download the videos along with the annotations.
-
-```shell
-bash download_data.sh
-```
+First of all, you have to visit the official [website](http://moments.csail.mit.edu/), fill in an application form for downloading the dataset. Then you will get the download link. You can use `bash preprocess_data.sh` to prepare annotations and videos. However, the download command is missing in that script. Remember to download the dataset to the proper place follow the comment in this script.
```
SuX97, line 23: First of all, you have to visit the official [website](http://moments.csail.mit.edu/), fill in an application form for downloading the dataset. Then you will get the download link. You can use ~~`bash preprocess_data.sh`~~`bash preprocess_data.sh $download_link` to prepare annotations and videos. ~~However, the download command is missing in that script. Remember to download the dataset to the proper place follow the comment in this script.~~

SuX97: Let's just keep it simple.

SuX97 (commented):
As suggested, MiT-v2 has changed its file structure and the preprocessing scripts afterward may fail. Can you please try applying for the new version? My application hasn't been answered for days.

tools/data/mit/preprocess_data.sh
kennymckormick: I'm not sure if the url can be downloaded using wget. For example, the google drive url can not be downloaded directly.

tools/data/mit/README.md
kennymckormick: The same concern as above.

tools/data/mit/preprocess_data.sh
SuX97: Not sure, either. Let's block this til I got the URL from the official site?

tools/data/mmit/preprocess_data.sh
```diff
@@ -7,8 +7,7 @@ fi
 
 cd ${DATA_DIR}
 
-wget -c https://www.dropbox.com/s/sz3yd1o0gf09amh/Multi_Moments_in_Time.zip?dl=0
-
+# Download the Multi_Moments_in_Time_Raw.zip here manually
 unzip Multi_Moments_in_Time_Raw.zip
 rm Multi_Moments_in_Time.zip
```
innerlee, line 12: BTW I feel that `rm` here is dangerous.

Say, if user happens to not have unzip installed on the machine, then no files were extracted, and the raw data is also gone.

kennymckormick: You are right. So I decide to remove the rm

tools/data/mit/preprocess_data.sh
kennymckormick: I found it's OK to use wget.

innerlee:
Okay, I wil merge this.

@dreamerlin  we may do another pass to remove the `rm` commands in bash scripts.

innerlee (approved):

PR #540, opened 2021-01-14
[docs] add config tag

congee524:


configs/detection/ava/README.md
```diff
@@ -6,7 +6,9 @@
 
 ## Introduction
 
-```
+[ALGORITHM]
+
+```citation
```
innerlee, line 11: is `citation` valid?

innerlee:
backbones are missing

configs/detection/ava/README.md
congee524: have been fixed

innerlee:
Pls also update the stats.py

congee524:
> backbones are missing

we don't have backbones. The current statistics are based on each folder.

C3D and I3D is only applied in video understanding, they are considered as algorithms.

innerlee:
What happens to papers with ` " " `

image

congee524:
> What happens to papers with `" "`
> 
> image

Since each model can only be counted for one paper, the operators and datasets involved in the model implementation should not be counted, so double quotation marks are added to ignore it.

congee524:
> Pls also update the stats.py

image

congee524:
image

innerlee (approved):

PR #547, opened 2021-01-17
[Feature] Demo stdet

kennymckormick:
Add demo for Spatio-Temporal Action detection.


demo/README.md
```diff
@@ -36,19 +36,19 @@ or use checkpoint url from `configs/` to directly load corresponding checkpoint,
 1. Recognize a video file as input by using a TSN model on cuda by default.
 
     ```shell
-    # The demo.mp4 and label_map.txt are both from Kinetics-400
+    # The demo.mp4 and k400_label_map.txt are both from Kinetics-400
```
innerlee, line 39: `k400_label_map` -> `label_map_k400`, so that label map files of different datasets are grouped together naturally

demo/demo_stdet.py
```diff
@@ -0,0 +1,327 @@
+import argparse
```
innerlee, line 1: The term`stdet` is cryptic. How about the full name `demo_spatial_temporal_det`

setup.cfg
```diff
@@ -19,6 +19,6 @@ line_length = 79
 multi_line_output = 0
 known_standard_library = pkg_resources,setuptools
 known_first_party = mmaction
-known_third_party = cv2,joblib,matplotlib,mmcv,numpy,pandas,pytest,scipy,seaborn,titlecase,torch
+known_third_party = cv2,joblib,matplotlib,mmcv,moviepy,numpy,pandas,pytest,scipy,seaborn,titlecase,torch,tqdm
```
innerlee, line 22: Is it possible to remove moviepy and tqdm?

innerlee:
Some minor comments

demo/demo_stdet.py
```diff
@@ -0,0 +1,327 @@
...
+                    diag0 = (location[0] + textwidth, location[1] - 14)
+                    diag1 = (location[0], location[1] + 2)
+                    cv2.rectangle(frame, diag0, diag1, plate[k + 1], -1)
+                    cv2.putText(frame, text, location, FONTFACE, FONTSCALE,
+                                FONTCOLOR, THICKNESS, LINETYPE)
+
+    return frames_
+
+
+# We only accept videos with regular fps (30 fps) now.
```
innerlee, line 86: Reason for this fps restriction?

demo/demo_stdet.py
```diff
@@ -0,0 +1,327 @@
...
+        help='give out a prediction per n frames')
+    parser.add_argument(
+        '--output-stepsize',
+        default=4,
+        type=int,
+        help=('show one frame per n frames in the demo, we should have: '
+              'predict_stepsize % output_stepsize == 0'))
+    parser.add_argument(
+        '--output-fps',
+        default=6,
```
innerlee, line 144: can this number be calculated automatically?

setup.cfg
innerlee: Okay we can leave tqdm since it is small and has no further dependency.

demo/demo_stdet.py
```diff
@@ -0,0 +1,327 @@
...
+        type=int,
+        help='the fps of demo video output')
+    args = parser.parse_args()
+    return args
+
+
+def frame_extraction(video_path):
+    # Load the video, extract frames into /tmp/video_name
+    target_dir = osp.join('/tmp', osp.basename(osp.splitext(video_path)[0]))
```
innerlee, line 153: How about windows?

demo/demo_stdet.py
```diff
@@ -0,0 +1,327 @@
...
+                        prediction[j].append((label_map[i + 1], result[i][j,
+                                                                          4]))
+            predictions.append(prediction)
+
+    results = []
+    for human_detection, prediction in zip(human_detections, predictions):
+        results.append(pack_result(human_detection, prediction, new_h, new_w))
+
+    def dense_timestamps(timestamps, n):
+        """Make it 3x frames."""
```
innerlee, line 307: 3x?

setup.cfg
kennymckormick: OK, I will remove moviepy

innerlee: Hmm mmcv has progressbar, let's use it :)

https://mmcv.readthedocs.io/en/latest/utils.html#progressbar

demo/demo_stdet.py
kennymckormick: It's also OK to create a temp directory under the current dir.

demo/demo_stdet.py
kennymckormick: Typo, it should be n x

demo/demo_stdet.py
kennymckormick: Because we use input like 8(clip_len)x8(frame_interval). If the fps is too low (10fps or lower), the input might be too dense. However, since most videos are 24, 25 or 30 fps, I think we may not suffer from this problem. In other words, this line can be removed I think,

demo/demo_stdet.py
kennymckormick: I think it would be better to let users decide the output fps they want. We just provide a reasonable set of args as default value.

innerlee (approved):

PR #556, opened 2021-01-21
Fix BSN pem_low_temporal_iou_threshold

SuX97:
In the paper:
image

So, the basic idea here is to resample pos/neg proposals to control the pos/neg ratio, the way the original repo & mma2 implement it is by using a mask applied on the loss:

```
        temporal_iou_loss = torch.sum(
            temporal_iou_loss *
            temporal_iou_weights) / torch.sum(temporal_iou_weights)
```
So basically this `temporal_iou_weights` masks out excessive neg samples to reach 1:2.

However in mma2 and original repo, the var `pem_low_temporal_iou_threshold` is set to 2.2 rather than 0.3 in the paper. This will result in `u_lmask` to be all zero, which makes no sense.

I suspect that it is a typo, `2.2` is fixed to `0.2`, still it is not consistent with the paper.

More experiments should be conducted, such as counting the neg/pos numbers.


SuX97:
The tensorflow implementation shows that it is a typo:

https://github.com/wzmsltw/BSN-boundary-sensitive-network/blob/master/PEM_train.py#L19

innerlee:
Nice finding. Any numbers/ckpts need to be updated?

innerlee:
Or we first update the code, and raise an issue to track the modelzoo updates?

SuX97:
@innerlee :
I have tested directly training with this change, the result:
auc: 66.4269
AR@1: 0.3280
AR@5: 0.4830
AR@10: 0.5532
AR@100: 0.7462

innerlee:
The bottom line is, the result should be reproducible. So Either update the ckpt and the result, or update the ckpt, config, and result.

SuX97:
@innerlee , the checkpoint is ready to go.

innerlee (approved):

PR #433, opened 2020-12-11
[Refactor] Refactor unit test structure

congee524:
Refactor unittest structure according to design document

PR #558, opened 2021-01-22
Fix precise_bn unittest.

SuX97:


congee524:
the bug will be fixed in #433

kennymckormick:
(Not related to this PR) BTW, have you changed the priority of the precise_bn_hook? I think maybe changing it to the right priority may lead to performance improvement @SuX97

SuX97:
@kennymckormick , haven't tested yet. I think the difference will not be significant. Before changing, only the bn updating of the last epoch will be ignored, since the updating in epoch x will be saved by save_checkpoint of epoch x+1. I'll change it after the verification.

SuX97:
Duplicated PR.

PR #567, opened 2021-01-27
Refactor ava hook

### LINKED ISSUE #549: Refactor evaluation of ava
Currently `metrics` in ava evaluation are `'proposal', 'bbox'`, which are not metrics. https://github.com/open-mmlab/mmaction2/blob/master/mmaction/datasets/ava_dataset.py#L319 .

We need to change metrics to `mAP, Recall` etc, and make `proposal/bbox` as metric_options

kennymckormick:
fix #549

Migration note :

You need to use `--eval mAP` during AVA models testing now.

AVADataset evaluation metrics modification: 
Supported metrics before: `proposal`,  `bbox`
Supported metrics after: `mAP` (the same as `bbox` before), the old `proposal` is removed.


innerlee:
Still WIP? Are there any configs need to be changed? Also give good warnings to help user migrate

mmaction/datasets/ava_dataset.py
```diff
@@ -316,10 +316,13 @@ def dump_results(self, results, out):
...
     def evaluate(self,
                  results,
-                 metrics=('proposal', 'bbox'),
+                 metrics=('mAP'),
                  metric_options=None,
                  logger=None):
         # need to create a temp result file
+        assert len(metrics) == 1 and metrics[0] == 'mAP', (
+            'For evaluation on '
+            'AVADataset, you need to use metrics "mAP"')
```
innerlee, line 325: add this:
`See https://github.com/open-mmlab/mmaction2/pull/567 for more info.`

innerlee:
It would be great to add some migration notes in the top post, to give detailed info on what has been changed, and how to change configs/cmds to the new way. Example: https://github.com/open-mmlab/mmpose/pull/382

innerlee:
Do you need to update modelzoo, if there are related metrics affected?

SuX97 (approved):
Just changing names looks good to me. But how do we evaluate AR@AN and Recall?

Can metrics be `['mAP', 'AR@AN', 'Recall']` since those metrics have been calculated anyway?

PR #574, opened 2021-01-28
[Fix] Fix a bug about inference config.

irvingzhang0512:
When running `./tools/dist_test.sh` with inference config, an `AttributeError` occurs.

```shell
  File "./tools/test.py", line 203, in <module>
    main()
  File "./tools/test.py", line 156, in main
    init_dist(args.launcher, **cfg.dist_params)
  File "/home/ubuntu/anaconda3/envs/zyy_pytorch1.6/lib/python3.7/site-packages/mmcv/utils/config.py", line 398, in __getattr__
    return getattr(self._cfg_dict, name)
  File "/home/ubuntu/anaconda3/envs/zyy_pytorch1.6/lib/python3.7/site-packages/mmcv/utils/config.py", line 43, in __getattr__
    raise ex
AttributeError: 'ConfigDict' object has no attribute 'dist_params'
```

Apparently, `dist_params` is required for `test.py`

### DIFF
--- a/configs/recognition/i3d/i3d_r50_video_inference_32x2x1_100e_kinetics400_rgb.py
+++ b/configs/recognition/i3d/i3d_r50_video_inference_32x2x1_100e_kinetics400_rgb.py
@@ -28,3 +28,6 @@ data = dict(
         ann_file=None,
         data_prefix=None,
         pipeline=test_pipeline))
+
+# runtime settings
+dist_params = dict(backend='nccl')
--- a/configs/recognition/r2plus1d/r2plus1d_r34_video_inference_8x8x1_180e_kinetics400_rgb.py
+++ b/configs/recognition/r2plus1d/r2plus1d_r34_video_inference_8x8x1_180e_kinetics400_rgb.py
@@ -33,3 +33,6 @@ data = dict(
         ann_file=None,
         data_prefix=None,
         pipeline=test_pipeline))
+
+# runtime settings
+dist_params = dict(backend='nccl')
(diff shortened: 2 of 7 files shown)


kennymckormick:
The inference config is only used by `demo.py`. So I think there is no need to add `dist_params` config. 
BTW, do we really need a separate inference config file, can't we use regular config files for `demo.py` (maybe with some modification)? @dreamerlin

congee524:
`dist_params` is defined in `configs/__base__/default_runtime.py`，you can print the config using `tools/analysis/print_config.py`. It has the property `dist_params` actually.

irvingzhang0512:
I thought inference configs were designed for test script. 

When testing with more crops and clips(e.g. ThreeCrop/TenCrop), sometimes maybe we need smaller `videos_per_gpu` than that in training because of gpu memory.

If it's designed for demo, no need to add this.

congee524:
> I thought inference configs were designed for test script.
> 
> When testing with more crops and clips(e.g. ThreeCrop/TenCrop), sometimes maybe we need smaller `videos_per_gpu` than that in training because of gpu memory.
> 
> If it's designed for demo, no need to add this.

we can use  `test_dataloader=dict(videos_per_gpu=2)` to change the batch size when testing.

irvingzhang0512:
> > I thought inference configs were designed for test script.
> > When testing with more crops and clips(e.g. ThreeCrop/TenCrop), sometimes maybe we need smaller `videos_per_gpu` than that in training because of gpu memory.
> > If it's designed for demo, no need to add this.
> 
> we can use `test_dataloader=dict(videos_per_gpu=2)` to change the batch size when testing.

Thanks for the tip

PR #582, opened 2021-02-01
Fix #565

### LINKED ISSUE #565: TSN temporal stream
Hello
I am trying to create a temporal stream for TSN using optical flow. Given the configuration below, I get the following error: `RuntimeError: Given groups=1, weight of size [64, 3, 7, 7], expected input[3, 10, 224, 224] to have 3 channels, but got 10 channels instead`.
The number of channels in the input after going through the training pipeline doesn't match the expected input of the first convolutional layer of the Resnet backbone model.
If I attempt to change the number of channels the backbone model expects, it leads to a failure when the model attempts to load the weights from torchvision.
What can be done so that I can train a TSN temporal stream?

You can find my Python configuration file as a text file here [tsn_r50_1x1x3_2e_ucf101_flow.py.txt](https://github.com/open-mmlab/mmaction2/files/5874657/tsn_r50_1x1x3_2e_ucf101_flow.py.txt)

kennymckormick:


mmaction/models/backbones/resnet.py
```diff
@@ -432,13 +432,15 @@ def _load_conv_params(conv, state_dict_tv, module_name_tv,
         """
 
         weight_tv_name = module_name_tv + '.weight'
-        conv.weight.data.copy_(state_dict_tv[weight_tv_name])
-        loaded_param_names.append(weight_tv_name)
+        if conv.weight.data.shape == state_dict_tv[weight_tv_name].shape:
```
innerlee, line 435: do we need to issue some warning if the shapes are mismatched?

kennymckormick: I think no warning is OK, since: 
1. The shape mismatch may only happen when in_channels != 3. In this case, the shape mismatch is not **unexpected**. Users assume that pre-train weights are for other modules (except conv1).
2. There is already a warning message in `_load_torchvision_checkpoint`, which indicates that parameters of `conv1` and `bn1` are not loaded.

innerlee: Its okay if pytorch could warn the user. As long as there are warnings, I'm ok with the change.

innerlee:
Changes are not covered by tests

kennymckormick:
> Changes are not covered by tests

I think it is covered: I add a testing case in which the ResNet is initialized with `in_channels=10` and use `torchvision://resnet50` for initialization.

innerlee (approved):

innerlee:
Then there must be tracking issues in Codcov

congee524:
`resnet` is been tested in `tests/test_models/test_common_modules/test_resnet.py`.

In our unit test refactor doc, `resnet` is considered as `common module`

kennymckormick:
> `resnet` is been tested in `tests/test_models/test_common_modules/test_resnet.py`.
> 
> In our unit test refactor doc, `resnet` is considered as `common module`

Sorry I didn't see it before,  I will reorganize related unittests.

congee524 (approved):

PR #586, opened 2021-02-02
[Feature] support gpu normalize

### LINKED ISSUE #470: GPU normalize
GPU normalization could speed up things greatly.

congee524:
fix #470


mmaction/datasets/pipelines/formating.py
```diff
@@ -276,6 +276,7 @@ def __call__(self, results):
             results (dict): The resulting dict to be modified and passed
                 to the next transform in pipeline.
         """
+        results['imgs'] = np.array(results['imgs'], dtype=np.float32)
```
dreamerlin, line 279: It is better to use an `if` to judge this

mmaction/utils/module_hooks.py
```diff
@@ -0,0 +1,52 @@
...
+        Module.register_forward_pre_hook(
+            build_from_cfg(module_hook_cfg, MODULE_HOOKS).hook_func())
+
+
+@MODULE_HOOKS.register_module()
+class GpuNormalize:
+
+    def __init__(self,
+                 input_format,
+                 mean=[123.675, 116.28, 103.53],
```
dreamerlin, line 20: use `tuple` as default

congee524:
@innerlee kindly ping :p

innerlee:
I suppose there should be more reviewers? cc. @hellock

configs/recognition/c3d/c3d_sports1m_gpu_normalize_16x1x1_45e_ucf101_rgb.py
```diff
@@ -0,0 +1,104 @@
+_base_ = '../../_base_/models/c3d_sports1m_pretrained.py'
+
+module_hooks = [
+    dict(
+        type='GPUNormalize',
+        hook_pos='forward_pre',
+        input_format='NCTHW',
+        mean=(104, 117, 128),
+        std=(1, 1, 1))
```
innerlee, line 9: Where do the values of mean and std come from?

congee524: This is the average value of each channel of all ucf101 images used by the author in the original paper. We also used this parameter to train the model before.

innerlee: ok, the std is not correct obviously

mmaction/datasets/pipelines/formating.py
```diff
@@ -276,6 +276,8 @@ def __call__(self, results):
             results (dict): The resulting dict to be modified and passed
                 to the next transform in pipeline.
         """
+        if not isinstance(results['imgs'], np.ndarray):
+            results['imgs'] = np.array(results['imgs'], dtype=np.float32)
```
innerlee, line 280: why float32?

mmaction/utils/module_hooks.py
```diff
@@ -0,0 +1,70 @@
...
+    Args:
+        mean (Sequence[float]): Mean values of different channels.
+        std (Sequence[float]): Std values of different channels.
+    """
+
+    def __init__(self, input_format, mean, std):
+        if input_format not in ['NCTHW', 'NCHW', 'NCHW_Flow', 'NPTCHW']:
+            raise ValueError(f'The input format {input_format} is invalid.')
+        self.input_format = input_format
+        _mean = torch.tensor(mean)
```
innerlee, line 44: does this tensor live on gpu?

innerlee: ok, not yet

mmaction/utils/module_hooks.py
```diff
@@ -0,0 +1,70 @@
...
+            self._std = _std[None, :, None, None]
+        elif input_format == 'NPTCHW':
+            self._mean = _mean[None, None, None, :, None, None]
+            self._std = _std[None, None, None, :, None, None]
+
+    def hook_func(self):
+
+        def normalize_hook(Module, input):
+            x = input[0]
+            if not hasattr(normalize_hook, 'mean'):
```
innerlee, line 63: Does this work with DataParallel?

congee524, line 63: typo, should be
`if not hasattr(self, 'mean'):`

congee524: yeah, the two values will be copied in each GPU. Have confirmed it by `print(self.mean.device)`

congee524:
Test GPU Normalize on 1080TI, TITAN XP, and Tesla V100.
Only on V100, GPU Normalize can significantly increase the training speed and shorten the training time by nearly half.

innerlee:
Okay, you can train a full model in V100, and upload the ckpt to modelzoo for future reference.

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -1202,8 +1202,7 @@ def __call__(self, results):
         modality = results['modality']
 
         if modality == 'RGB':
-            n = len(results['imgs'])
-            h, w, c = results['imgs'][0].shape
+            n, h, w, c = np.shape(results['imgs'])
```
innerlee, line 1205: np.shape implicitly calls `asarray`. I'm not sure if this induces any data copy.

Could you pls comapre the speed of `np.shape(results['imgs'])` and the original version?

mmaction/utils/module_hooks.py
```diff
@@ -0,0 +1,75 @@
...
+            self._std = _std[None, :, None, None]
+        elif input_format == 'NPTCHW':
+            self._mean = _mean[None, None, None, :, None, None]
+            self._std = _std[None, None, None, :, None, None]
+
+    def hook_func(self):
+
+        def normalize_hook(Module, input):
+            x = input[0]
+            if not hasattr(self, 'mean'):
```
innerlee, line 63: Have you tested on the normal, **non**-distribued DataParallel?

congee524: I guess this is doesn't work with non-distributed DataParallel since the module data only stored in cuda:0.

Actually, MMDataParallel has not supported training on multi-GPUs. Do we still need to test it in MMDataParallel?
image

mmaction/datasets/pipelines/augmentations.py
congee524: I've rolled back， and `np.shape` is very slow indeed.

mmaction/utils/module_hooks.py
innerlee: If we do not cache the tensor in self.mean and self.std, and call the `.to(device)` every time, will the speed be affected?

congee524: There are no free GPUs at the moment. I'll see the result later

congee524: The speed is almost unaffected

mmaction/utils/module_hooks.py
```diff
@@ -0,0 +1,75 @@
...
+            self._std = _std[None, :, None, None, None]
+        elif input_format == 'NCHW':
+            self._mean = _mean[None, :, None, None]
+            self._std = _std[None, :, None, None]
+        elif input_format == 'NCHW_Flow':
+            self._mean = _mean[None, :, None, None]
+            self._std = _std[None, :, None, None]
+        elif input_format == 'NPTCHW':
+            self._mean = _mean[None, None, None, :, None, None]
+            self._std = _std[None, None, None, :, None, None]
```
innerlee, line 57: add an else branch

innerlee (approved):

PR #587, opened 2021-02-02
link CONTRIBUTING.md to the one in MMCV

dreamerlin:


innerlee:
The presence of the file actually has some effect. Removing it would defeat the purpose

image

See https://docs.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors#adding-a-contributing-file

dreamerlin:
> The presence of the file actually has some effect. Removing it would defeat the purpose
> 
> image
> 
> See https://docs.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors#adding-a-contributing-file

Truly, what do you think @hellock

innerlee (approved):

PR #589, opened 2021-02-03
Fix ANet configs setting and related results

kennymckormick:
For using the weight of a **Recognizer** for initialization. We should use the `load_from` args instead of writing `pretrained=blahblah` in the backbone config.


kennymckormick:
Can load ckpt properly now. @innerlee

innerlee:
Please explain the reason for the change at the top post :)

configs/recognition/tsn/tsn_r50_320p_1x1x8_150e_activitynet_video_flow.py
```diff
@@ -67,7 +65,6 @@
         ann_file=ann_file_train,
         data_prefix=data_root,
         filename_tmpl='flow_{}_{:05d}.jpg',
-        with_offset=True,
```
innerlee, line 70: What's the effect of removing this line?

kennymckormick: When training with `video` file lists, it should be set as False (or it will not run successfully).

innerlee (approved):

PR #600, opened 2021-02-05
[Improvement] Modify default value of `save_best`

### LINKED ISSUE #599: About default value of `save_best` in EvalHook
By default `save_best=None`, which disables saving best eval ckpt.

I think saving best ckpt is a basic function during training. It doesn't require a lot of resources and should be enabled by default.

It could be done by either set `save_best='auto'` or add `evalution=dict(save_best='xxx')` in all configs. I prefer the latter.

irvingzhang0512:
Fix #599

### DIFF
--- a/mmaction/core/evaluation/eval_hooks.py
+++ b/mmaction/core/evaluation/eval_hooks.py
@@ -60,7 +60,7 @@ class EvalHook(Hook):
                  start=None,
                  interval=1,
                  by_epoch=True,
-                 save_best=None,
+                 save_best='auto',
                  rule=None,
                  **eval_kwargs):
         if 'key_indicator' in eval_kwargs:
--- a/mmaction/core/evaluation/eval_hooks.py
+++ b/mmaction/core/evaluation/eval_hooks.py
@@ -35,12 +35,10 @@ class EvalHook(Hook):
             ``mean_average_precision``, ``mmit_mean_average_precision``
             for action recognition dataset (RawframeDataset and VideoDataset).
             ``AR@AN``, ``auc`` for action localization dataset.
-            (ActivityNetDataset). ``Recall@0.5@100``, ``AR@100``,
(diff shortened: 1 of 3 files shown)


irvingzhang0512:
if `save_best='auto'`, then we have to set `rule` manually. Any suggestions? @dreamerlin

dreamerlin:
I guess you mean the case `rule=None` & `save_best='auto`.  Actually in this case, `self._init_rule` will be called twice.

At first time: https://github.com/open-mmlab/mmaction2/blob/master/mmaction/core/evaluation/eval_hooks.py#L97-L99

The result is: `self.rule=None`, `self.key_indicator='auto'`, `self.compare_func` is not set

At second time: https://github.com/open-mmlab/mmaction2/blob/master/mmaction/core/evaluation/eval_hooks.py#L245-L247

This time, the `key_indicator` is not 'auto', and then the result: `self.rule` is not None but between 'greater' and 'less', self.key_indicator is not 'auto' and `se;f.compare_func` is determined.

dreamerlin:
You may also need to change default value of `DisyEvalHook`. And since you are here, Coulu you also help to change the default of docstring ?

irvingzhang0512:
Thanks for your clarification. It turns out that test failure is caused by `Mock`. Tests, Docs & default values are updated.

BTW, do we still need `The interval of EvalHook should be divisible by that of CheckpointHook` in the docstring? I thought it was designed for symlink

dreamerlin:
> Thanks for your clarification. It turns out that test failure is caused by `Mock`. Tests, Docs & default values are updated.
> 
> BTW, do we still need `The interval of EvalHook should be divisible by that of CheckpointHook` in the docstring? I thought it was designed for symlink

Yeah, it is outdated, thanks for your correction!

dreamerlin (approved):

PR #608, opened 2021-02-09
[Feature] Add `--test-last` & `--test-best` for `tools/train.py` to test checkpoints after training 

kennymckormick:


irvingzhang0512:
Advice: Maybe we should test the best ckpt(if exsits) saved by eval hook.

innerlee:
What's the difference between this test and the test.py?

kennymckormick:
> What's the difference between this test and the test.py?

They are the same, just don't want to run `python test.py` manually after training.

innerlee:
The implementation is too brute-force. Do you have more elegent way to do this? 
Hooks? workflow?

kennymckormick:
> The implementation is too brute-force. Do you have more elegent way to do this?
> Hooks? workflow?

What's the motivation? The feature is straightforward: run testing for the last ckpt (usually the best one) after training and it will only run once. What's the drawback of current design?

innerlee:
@dreamerlin @SuX97 what do you think?

SuX97:
> @dreamerlin @SuX97 what do you think?

I think it is OK but maybe add some options such as setting the saved checkpoint to be the best or the latest.

Hardcoding the result as `final_pred.pkl` seems brute-force and can cause troubles such as overwriting.

Still, I prefer to keep the API simple but scripts/tools complicated. Such as a new script called `slurm_train_test.sh`.

dreamerlin:
I vote for test at the end of training, since it can save time to launch a test script

innerlee:
Okay, there are more votes in favor the the feature. So @kennymckormick may adjust the code incorprating @SuX97 's suggestion, and then it is good to go.

mmaction/apis/train.py
```diff
@@ -154,3 +157,27 @@ def train_model(model,
...
+        test_dataset = build_dataset(cfg.data.test, dict(test_mode=True))
+        dataloader_setting = dict(
+            videos_per_gpu=1,
+            workers_per_gpu=4,
+            num_gpus=len(cfg.gpu_ids),
+            dist=distributed,
+            shuffle=False)
+        test_dataloader = build_dataloader(test_dataset, **dataloader_setting)
+        outputs = multi_gpu_test(model, test_dataloader,
+                                 osp.join(cfg.work_dir, 'tmp'), False)
```
dreamerlin, line 171:
```python
gpu_collect = cfg.get('evaluation', {}).get('gpu_collect', False)
tmpdir= cfg.get('evaluation', {}).get('tmpdir', osp.join(cfg.work_dir, 'tmp'))
dataloader_setting = dict(
    videos_per_gpu=cfg.data.get('videos_per_gpu', 1),
    workers_per_gpu=cfg.data.get('workers_per_gpu', 1),
    num_gpus=len(cfg.gpu_ids),
    dist=distributed,
    shuffle=False)
dataloader_setting = dict(dataloader_setting, **cfg.data.get('test_dataloader', {}))
test_dataloader = build_dataloader(test_dataset, **dataloader_setting)
outputs = multi_gpu_test(model, test_dataloader, tmpdir, gpu_collect)
```

mmaction/apis/train.py
```diff
@@ -154,3 +157,27 @@ def train_model(model,
...
+        rank, _ = get_dist_info()
+        if rank == 0:
+            out = osp.join(cfg.work_dir, 'final_pred.pkl')
+            test_dataset.dump_results(outputs, out)
+
+            eval_cfg = cfg.get('evaluation', {})
+            if 'interval' in eval_cfg:
+                eval_cfg.pop('interval')
```
dreamerlin, line 179:
```python
  for key in [
          'interval', 'tmpdir', 'start', 'gpu_collect', 'save_best', 'rule',
          'by_epoch'
  ]:
      eval_kwargs.pop(key, None)
```

docs/getting_started.md
```diff
@@ -343,6 +343,7 @@ If you want to specify the working directory in the command, you can add an argu
 Optional arguments are:
 
 - `--validate` (**strongly recommended**): Perform evaluation at every k (default value is 5, which can be modified by changing the `interval` value in `evaluation` dict in each config file) epochs during the training.
+- `--test`: Test the final checkpoint when training is over, save the prediction to `${WORK_DIR}/final_pred.pkl`
```
congee524, line 346: Add the arguments in docs_zh_CN/getting_started.md

mmaction/apis/train.py
```diff
@@ -1,22 +1,25 @@
...
 from ..utils import PreciseBNHook, get_root_logger
+from .test import multi_gpu_test
 
 
 def train_model(model,
                 dataset,
                 cfg,
                 distributed=False,
                 validate=False,
+                test=False,
```
congee524, line 22: rename it with `test_model_after_train` may be better.

mmaction/apis/train.py
```diff
@@ -154,3 +161,65 @@ def train_model(model,
     if cfg.omnisource:
         runner_kwargs = dict(train_ratio=train_ratio)
     runner.run(data_loaders, cfg.workflow, cfg.total_epochs, **runner_kwargs)
+
+    if test['test_last'] or test['test_best']:
```
dreamerlin, line 165: Actually, https://github.com/open-mmlab/mmaction2/blob/master/mmaction/core/evaluation/eval_hooks.py#L225, we can use `save_best= cfg.get('evaluation', {}).get('save_best', False)` as an indicator to judge whether to test the best ckpt, `runner.meta['hook_msgs']['best_ckpt']` or `eval_hook.best_ckpt_path` can be use as the path of the best ckpt.

mmaction/apis/train.py
kennymckormick: Thanks for the advise

mmaction/apis/train.py
kennymckormick: Thanks for the advise

mmaction/apis/train.py
```diff
@@ -154,3 +161,65 @@ def train_model(model,
...
+
+                eval_cfg = cfg.get('evaluation', {})
+                for key in [
+                        'interval', 'tmpdir', 'start', 'gpu_collect',
+                        'save_best', 'rule', 'by_epoch', 'broadcast_bn_buffers'
+                ]:
+                    eval_cfg.pop(key, None)
+
+                eval_res = test_dataset.evaluate(outputs, **eval_cfg)
+                runner.logger.info('Testing results of the last checkpoint')
```
dreamerlin, line 223: May need to log best and last respectively

kennymckormick: Yes, I will modify here.

mmaction/apis/train.py
kennymckormick: For the first suggestion, I think `save_best` doesn't necessarily mean the user also wants to test the best (during validation) checkpoint, so I let users specify it explicitly. 
For the second one, I think that makes sense, I will use `eval_hook.best_ckpt_path` and check if it exists.

mmaction/apis/train.py
kennymckormick: I think using the name `test` is of the same style with `validate`.

mmaction/apis/train.py
dreamerlin: Agree, if the test case is the same as the validation case, it is indeed no need to test the best one again.

mmaction/apis/train.py
```diff
@@ -154,3 +160,64 @@ def train_model(model,
...
+    if test['test_last'] or test['test_best']:
+        best_ckpt_path = None
+        if test['test_best']:
+            if hasattr(eval_hook, 'best_ckpt_path'):
+                best_ckpt_path = eval_hook.best_ckpt_path
+
+            if best_ckpt_path is None or not osp.exists(best_ckpt_path):
+                test['test_best'] = False
+                runner.logger.info('Warning: test_best set as True, but is '
+                                   'not applicable')
```
congee524, line 173: Maybe more info should be given, such as set save_best in evaluation.

mmaction/apis/train.py
```diff
@@ -154,3 +160,64 @@ def train_model(model,
...
+
+        if not (test['test_last'] or test['test_best']):
+            return
```
congee524, line 176:
```python
if best_ckpt_path is None or not osp.exists(best_ckpt_path):
                test['test_best'] = False
                runner.logger.info('Warning: test_best set as True, but is '
                                   'not applicable')
                if not test['test_last']:
                    return
```

mmaction/apis/train.py
dreamerlin: default value for `test` should be `dict`

kennymckormick:
comments resolved

dreamerlin:
LGTM if it runs successfully.

dreamerlin (approved):

innerlee:
CI fails

innerlee:
CI is fixed in master. Pls rebase~

PR #623, opened 2021-02-21
Add out_dir to EvalHook

dumbPy:
Add `out_dir` to `EvalHook` so that the evaluation results can be dumped into a directory for further analysis like plotting confusion matrix.

Depending on automated comparison of a single metric is not always enough. This helps me plot my own graphs and confusion matrices.

### DIFF
--- a/mmaction/core/evaluation/eval_hooks.py
+++ b/mmaction/core/evaluation/eval_hooks.py
@@ -44,6 +44,9 @@ class EvalHook(Hook):
             .etc will be inferred by 'greater' rule. Keys contain 'loss' will
             be inferred by 'less' rule. Options are 'greater', 'less', None.
             Default: None.
+        out_dir (str | None, optional): Output Directory to dump evaluation
+            results.
+            Default: None.
         **eval_kwargs: Evaluation arguments fed into the evaluate function of
             the dataset.
     """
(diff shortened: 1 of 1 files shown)


mmaction/core/evaluation/eval_hooks.py
```diff
@@ -60,6 +63,7 @@ def __init__(self,
                  by_epoch=True,
                  save_best='auto',
                  rule=None,
+                 out_dir=None,
```
innerlee, line 66: How about saving it to work_dir, like the ckpts?

dreamerlin: Also vote for saving to the `work_dir`

kennymckormick: An advice: you can set the default value of `out_dir` as `work_dir`. For example, if  `out_dir is None`, you output the eval results to `work_dir`.

kennymckormick: It's OK to save eval results by default, since it doesn't take much storage space.

innerlee: Okay

dreamerlin:
`EvalHook` has been moved to MMCV, so we will close this PR

PR #624, opened 2021-02-22
[Improvement] Add warnings of uninstalling mmdet and supplementary documents

### LINKED ISSUE #620: ValueError: FastRCNN is not registered in LOCALIZERS, RECOGNIZERS or DETECTORS
Hello
when I train AVA dataset find a error
ValueError: FastRCNN is not registered in LOCALIZERS, RECOGNIZERS or DETECTORS
how to solve it 
think you very much

congee524:
fix #620

### DIFF
--- a/mmaction/models/builder.py
+++ b/mmaction/models/builder.py
@@ -74,6 +74,10 @@ def build_model(cfg, train_cfg=None, test_cfg=None):
         return build_recognizer(cfg, train_cfg, test_cfg)
     if obj_type in DETECTORS:
         return build_detector(cfg, train_cfg, test_cfg)
+    class_in_mmdet = ['FastRCNN', 'MaxIoUAssignerAVA', 'RandomSampler']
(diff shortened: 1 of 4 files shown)


mmaction/models/builder.py
```diff
@@ -74,6 +74,10 @@ def build_model(cfg, train_cfg=None, test_cfg=None):
         return build_recognizer(cfg, train_cfg, test_cfg)
     if obj_type in DETECTORS:
         return build_detector(cfg, train_cfg, test_cfg)
+    class_in_mmdet = ['FastRCNN', 'MaxIoUAssignerAVA', 'RandomSampler']
```
innerlee, line 77: Does RandomSampler belong to model?

congee524: emmm, and `MaxIoUAssignerAVA` is also in mmaction2.

innerlee (approved):

PR #625, opened 2021-02-22
[Improvement] Support soft label for CrossEntropyLoss

irvingzhang0512:
## Motivation
Soft labels are required for data augmentation methods like mixup, cutmix, etc. For now, official `CrossEntropyLoss`  doesn't support soft labels.

## Details
+ If number of classes is 4.
    + Hard labels example: `[0, 1, 2]`
    + Corresponding soft labels are `[[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0]]`
    + Mixup/Cutmix may create soft labels like `[[0.8, 0.2, 0, 0], [0.1, 0.9, 0, 0], [0, 0, 0.6, 0.4]]`
+ It's difficult to get a proper name for this pr and related args.
    + `Sparse` is borrowed from TensorFlow and MXNet, TensorFlow has  `CategoricalCrossentropy`(accept non-sparse, one-hot like labels) and `SparseCategorialCrossentropy`(just like `CrossEntropyLoss` in pytorch).
    + BTW, **one-hot like labels** is more clear and meaningful for me.
+ For now, this pr detects loss type by the shape of label.

## TODO
+ [x] Modify `CrossEntropyLoss` in `mmaction/models/losses/cross_entropy_loss.py`.
+ [x] unittest
+ [x]  changelog.

### DIFF
--- a/mmaction/models/losses/cross_entropy_loss.py
+++ b/mmaction/models/losses/cross_entropy_loss.py
@@ -44,18 +44,15 @@ class CrossEntropyLoss(BaseWeightedLoss):
             torch.Tensor: The returned CrossEntropy loss.
         """
         if self.sparse_label:
-            # sparse labels
             if self.class_weight is not None:
                 assert 'weight' not in kwargs, \
                     "The key 'weight' already exists."
                 kwargs['weight'] = self.class_weight.to(cls_score.device)
             loss_cls = F.cross_entropy(cls_score, label, **kwargs)
         else:
-            # non-sparse(one-hot like) labels
             assert cls_score.size() == label.size()
             assert cls_score.dim() == 2
 
-            # cal cross entropy loss with one-hot like labels
             lsm = F.log_softmax(cls_score, 1)
             if self.class_weight is not None:
                 lsm = lsm * self.class_weight.unsqueeze(0)
--- a/mmaction/models/losses/cross_entropy_loss.py
+++ b/mmaction/models/losses/cross_entropy_loss.py
@@ -9,6 +9,16 @@ from .base import BaseWeightedLoss
 class CrossEntropyLoss(BaseWeightedLoss):
     """Cross Entropy Loss.
 
+    Support two kinds of labels and their corresponding loss type. It's worth
+    mentioning that loss type will be detected by the shape of ``cls_score``
+    and ``label``.
+    1) Sparse label: integers in range [0, num_classes - 1].
+    2) Non-sparse label(one-hot like labels): floats in range [0, 1]. It shares
+        the same shape with ``cls_score``.
+
+    For example, if the number of classes is 4, sparse labels could be [0, 2]
+    and its corresponding non-sparse labels are [[1, 0, 0, 0], [0, 0, 1, 0]].
+
     Args:
         loss_weight (float): Factor scalar multiplied on the loss.
             Default: 1.0.
@@ -29,13 +39,7 @@ class CrossEntropyLoss(BaseWeightedLoss):
 
         Args:
             cls_score (torch.Tensor): The class score.
-            label (torch.Tensor): The ground truth label. Support two kinds of
-                label, sparse label and non-sparse label. Sparse labels are
-                integers in range [0, num_classes - 1]. Non-sparse labels, or
-                one-hot like labels, are floats in range [0, 1]. If the shape
-                of ``cls_score`` is [N, num_classes], the shape of ``label``
-                should be [N] for sparse labels and [N, num_classes] for
-                non-sparse labels.
+            label (torch.Tensor): The ground truth label.
             kwargs: Any keyword argument to be used to calculate
                 CrossEntropy loss.
 
@@ -43,6 +47,8 @@ class CrossEntropyLoss(BaseWeightedLoss):
             torch.Tensor: The returned CrossEntropy loss.
         """
         if cls_score.size() == label.size():
+            # cal loss for non-sparse(one-hot like) label
+
             assert cls_score.dim() == 2, 'Only support 2-dim non-sparse labels'
 
             lsm = F.log_softmax(cls_score, 1)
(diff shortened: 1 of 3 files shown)


mmaction/models/losses/cross_entropy_loss.py
```diff
@@ -16,13 +16,20 @@ class CrossEntropyLoss(BaseWeightedLoss):
             as None, use the same weight 1 for all classes. Only applies
             to CrossEntropyLoss and BCELossWithLogits (should not be set when
             using other losses). Default: None.
+        sparse_label (bool): Whether to use sparse labels or not. Sparse labels
```
innerlee, line 19: How about detecting which kind of loss from the shape of label?

irvingzhang0512: Sounds good. This could also simplify the codes of mixup/cutmix.

mmaction/models/losses/cross_entropy_loss.py
```diff
@@ -9,6 +9,16 @@
 class CrossEntropyLoss(BaseWeightedLoss):
     """Cross Entropy Loss.
 
+    Support two kinds of labels and their corresponding loss type. It's worth
+    mentioning that loss type will be detected by the shape of ``cls_score``
+    and ``label``.
+    1) Sparse label: integers in range [0, num_classes - 1].
+    2) Non-sparse label(one-hot like labels): floats in range [0, 1]. It shares
```
innerlee, line 16: What's the difference between soft-label and non-sparse label?

irvingzhang0512: I couldn't tell the difference between non-sparse label/soft-label/one-hot like label. I just don't know which term is more former and meaningful.

innerlee: let's use soft-label

mmaction/models/losses/cross_entropy_loss.py
```diff
@@ -9,6 +9,20 @@
 class CrossEntropyLoss(BaseWeightedLoss):
     """Cross Entropy Loss.
 
+    Support two kinds of labels and their corresponding loss type. It's worth
+    mentioning that loss type will be detected by the shape of ``cls_score``
+    and ``label``.
+    1) Sparse label: This label is an integer array and all of the elements are
```
innerlee, line 15: Sparse -> One-hot

mmaction/models/losses/cross_entropy_loss.py
```diff
@@ -36,10 +50,33 @@ def _forward(self, cls_score, label, **kwargs):
         Returns:
             torch.Tensor: The returned CrossEntropy loss.
         """
-        if self.class_weight is not None:
-            assert 'weight' not in kwargs, "The key 'weight' already exists."
-            kwargs['weight'] = self.class_weight.to(cls_score.device)
-        loss_cls = F.cross_entropy(cls_score, label, **kwargs)
+        if cls_score.size() == label.size():
+            # cal loss for soft label
```
innerlee, line 54: One thing to take care is the `kwargs`. Please check that no kwargs are ignored silently

mmaction/models/losses/cross_entropy_loss.py
irvingzhang0512: Non-sparse labels are one-hot like labels, aka a probability distribution.

Sparse label is a list of integers.

innerlee: I think this is a misuse of "sparse". How about just "labels"

irvingzhang0512: I'm not sure about the misuse thing.

[TF SparseCategoricalCrossentropy](https://www.tensorflow.org/api_docs/python/tf/keras/losses/SparseCategoricalCrossentropy)

[Book: Hands-On Machine Learning book with Scikit-Learn and TF](https://stats.stackexchange.com/questions/427748/what-does-it-mean-when-someone-says-they-have-sparse-labels)

[GluonCV](https://github.com/dmlc/gluon-cv/blob/2da042f8982b8c46cdf04b0646c845f1175402ef/gluoncv/loss.py#L275)

innerlee: It seems that the usage comes from tf. It is not good term, lets stop its popularization.

Okay joking :) I think "label" is clear enough here. If you want to differentiate, use "hard label"

mmaction/models/losses/cross_entropy_loss.py
```diff
@@ -9,6 +9,17 @@
 class CrossEntropyLoss(BaseWeightedLoss):
     """Cross Entropy Loss.
 
+    Support two kinds of labels and their corresponding loss type. It's worth
+    mentioning that loss type will be detected by the shape of ``cls_score``
+    and ``label``.
+    1) Hard label: This label is an integer array and all of the elements are
+        in the range [0, num_classes - 1]. This label's shape should be
+        ``cls_score``'s shape with the `num_classes` dimension removed.
+    2) Soft label(one-hot like label): This label is a probability distribution
```
innerlee, line 18: ~~(one-hot like label)~~ (using probablity distribution over classes as label)

`One-hot like` gives more confusion than clarification. If something resembles one-hot, maybe it is "Two-hot" or "One-hotter/cooler"?
"The label shape is the similar to one-hot encoding" is the last thing one may think of

mmaction/models/losses/cross_entropy_loss.py
innerlee: cal -> calculate

innerlee:
Some minor comments.

innerlee (approved):

kennymckormick:
Thanks for the PR~ Besides have u find some settings in which soft labels work in action recognition? I have tried to find one but failed. If you find such a setting, you can add the config file which uses soft label for training and achieves better results.

kennymckormick (approved):

irvingzhang0512:
> Thanks for the PR~ Besides have u find some settings in which soft labels work in action recognition? I have tried to find one but failed. If you find such a setting, you can add the config file which uses soft label for training and achieves better results.

I plan to implement mixup/cutmix in two weeks, which use soft label.
image

PR #648, opened 2021-02-25
Fix config refactor bug

### LINKED ISSUE #645: mmcv version
  File "tools/test.py", line 196, in <module>
    main()
  File "tools/test.py", line 134, in main
    if cfg.test_cfg is None:
  File "/home/cb/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmcv/utils/config.py", line 363, in __getattr__
    return getattr(self._cfg_dict, name)
  File "/home/cb/anaconda3/envs/open-mmlab/lib/python3.7/site-packages/mmcv/utils/config.py", line 43, in __getattr__
    raise ex
AttributeError: 'ConfigDict' object has no attribute 'test_cfg'

when i test model, above error happened, i try to uninstall mmcv and reinstall mmcv=1.1.6.  
but error also happend

congee524:
fix #645

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -141,7 +141,7 @@ def main():
         torch.backends.cudnn.benchmark = True
     cfg.data.test.test_mode = True
 
-    if cfg.model.get('test_cfg') is None and cfg.model.get('test_cfg') is None:
+    if cfg.model.get('test_cfg') is None and cfg.get('test_cfg') is None:
         cfg.model.setdefault('test_cfg',
                              dict(average_clips=args.average_clips))
     else:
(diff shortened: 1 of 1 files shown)


innerlee:
Do a global search to make sure that `cfg.test_cfg` does not occur anymore

congee524:
> Do a global search to make sure that `cfg.test_cfg` does not occur anymore

emm, I did a global search but didn't notice it. and only did the training model test.

There should be no problem now

innerlee:
How about `cfg.train_cfg`?

congee524:
> How about `cfg.train_cfg`?

have confirmed it.

innerlee (approved):

dreamerlin (approved):

PR #658, opened 2021-02-28
Add faq for how to fix stages of backbone when finetuning a model

dreamerlin:


innerlee:
The instruction tell users to modify codes by themselves. It is okay by itself.

Is it possible to support freezing stages for commonly used backbones?

dreamerlin:
> The instruction tell users to modify codes by themselves. It is okay by itself.
> 
> Is it possible to support freezing stages for commonly used backbones?

For current support model backbones except `C3D`, users can set `frozen_stages` to freeze stages. Actually, all backbone inherit from `ResNet` or `ResNet3D` support freeze stages

dreamerlin:
> > The instruction tell users to modify codes by themselves. It is okay by itself.
> > Is it possible to support freezing stages for commonly used backbones?
> 
> For current supported model backbones except `C3D`, users can set `frozen_stages` to freeze stages. Actually, all backbone inherit from `ResNet` or `ResNet3D` support freeze stages

MobileNetV2 also supports it

innerlee:
Cool then this faq could list the supported backbones as bonus info

innerlee (approved):

PR #668, opened 2021-03-02
add limiter for demo

dreamerlin:


demo/webcam_demo.py
```diff
@@ -43,14 +44,29 @@ def parse_args():
         type=int,
         default=1,
         help='number of latest clips to be averaged for prediction')
+    parser.add_argument(
+        '--drawing-fps',
+        type=int,
+        default=-1,
```
innerlee, line 50: We can make a reasonable default value, such as drawing fps = inference fps =  4

innerlee (approved):

PR #675, opened 2021-03-04
[Bug Fix] Add bn buffer sync in eval_hook

dreamerlin:
I modified the eval_hook.py to support broadcast the buffer of batch_norm in rank 0 to other ranks, so as to ensure that the test results in the log can be consistent with individual test results with the saved checkpoint.


dreamerlin:
SlowOnly has been benchmarked in this PR and achieves the same acc between evaluation and test. The result seems to be acceptable so the PR can be merged.

innerlee (approved):

PR #683, opened 2021-03-06
[Bug] Add missing 'loss_aux' and related unittest

dreamerlin:
'loss_aux' got lost in losses, this PR adds it back together with related unittest.


innerlee:
Is the ckpt affected?

dreamerlin:
> Is the ckpt affected?

No, we haven't release 2D TPN

dreamerlin:
And this changes only affect 2D TPN

innerlee (approved):

PR #686, opened 2021-03-07
[Fix] reverse key-value in anet filelist generation

dreamerlin:


innerlee (approved):

innerlee:
What's the effect of the bug?

dreamerlin:
> What's the effect of the bug?

Nothing. This PR is for those whose video filename is not totally same with the url, e.g. `v_abcde` vs `abcde`.

PR #693, opened 2021-03-08
[Fix] fix flow norm cfg typo

### LINKED ISSUE #692: Wrong Value for flow normalization
`img_norm_cfg` should be `dict(mean=[128, 128], std=[128,128])`, not imagenet one.

congee524:
fix #692


innerlee:
How to ensure that configs are valid?

congee524:
> How to ensure that configs are valid?

The modified?
Keep them consistent with the norm config in the log file of the `recognition/tsn/README.md`

congee524:
we convert flow_x flow_y to Single channel grayscale image and thus use `mean=[128, 128], std=[128, 128]`

innerlee:
> The modified?

No, the original incorrect one

congee524:
There was a typo when adding. Maybe the reviewer didn't notice it. After all, the training process has no problem.

congee524:
Can't find that PR. It's too long

innerlee (approved):

innerlee:
Note to future:
We want our config to be able to reproduce the results. Such typos should be avoided as possible.

PR #753, opened 2021-03-24
fix lint

dreamerlin:


docs_zh_CN/README.md
```diff
@@ -1 +1 @@
-../README_zh-CN.md
\ No newline at end of file
+../README_zh-CN.md
```
congee524: What's the different

innerlee: Yeah this file keeps changing. Very odd

innerlee (approved):

PR #781, opened 2021-04-02
[Improvement] Add softmax option for pytorch2onnx tool.

irvingzhang0512:
## Description
Now the exported recognizer onnx model doesn't have a softmax layer. In order to get predictions in range `[0, 1]`, `--softmax` option is added to `pytorch2onnx.py`, to add a softmax layer at the end of recognizers.

## usage

```shell
python tools/pytorch2onnx.py $CONFIG_PATH $CHECKPOINT_PATH --shape $SHAPE --verify --softmax
```

## TODO
+ [x] codes
+ [x] docs
+ [x] unittest
+ [x] changelog

## Question
Why does `forward_dummy` return a tuple `(outs, )` instead of `outs`?

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,19 @@
 ## Changelog
 
+### Master
+
+**Highlights**
+
+**New Features**
+
+**Improvements**
+
+- Add softmax option for pytorch2onnx tool ([#781](https://github.com/open-mmlab/mmaction2/pull/781))
+
+**Bug and Typo Fixes**
+
+**ModelZoo**
+
 ### 0.13.0 (31/03/2021)
 
 **Highlights**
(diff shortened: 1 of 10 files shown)


dreamerlin (approved):

kennymckormick (commented):
I'm not familiar with onnx and have a question: how will the forward function of an onnx model be used?  For example, for 2D recognizers, we should first take the average of scores of different segments, then apply a softmax to get the prob, will onnx models have the same behavior?

irvingzhang0512:
> I'm not familiar with onnx and have a question: how will the forward function of an onnx model be used? For example, for 2D recognizers, we should first take the average of scores of different segments, then apply a softmax to get the prob, will onnx models have the same behavior?

You can think of onnx models as static computation graph. Add `nn.functional.softmax` corresponds to add a softmax layer for the computation graph. The forward pass of onnx is the same as that of pytorch.
Technically, onnx is the frontend. It defines static computation graph(layer/weights/edge). Onnxruntime is the backend(takes processed images and static computation graph as inputs, do the actual model forward to get model result)

innerlee (approved):

innerlee:
Thanks!

PR #782, opened 2021-04-04
[Improvement] Refactor STDet

kennymckormick:


mmaction/datasets/pipelines/augmentations.py
```diff
@@ -351,6 +352,8 @@ def __init__(self, scales, mode='range', **kwargs):
                              f'but got {self.mode}')
         self.scales = scales
         self.kwargs = kwargs
+        warnings.warn('This augmentation is not used in our experiments, '
```
innerlee, line 355: Reason for this warn?

kennymckormick: This aug is not used & we don't know if it works properly in the pipeline.

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -391,132 +394,99 @@ def __repr__(self):
...
 
-        return results
-
-    def __repr__(self):
-        return f'{self.__class__.__name__}(scale_factor={self.scale_factor})'
-
-
-@PIPELINES.register_module()
-class EntityBoxCrop:
+def box_crop(results, crop_bbox):
```
innerlee, line 419: Since `EntityBoxCrop` etc. are exported as part of our interface, so we need to add deprecation warnings first, and then remove them.

You can make a dummy implementation of `EntityBoxCrop` (with dep warn) by using the new `box_crop` function.

mmaction/datasets/pipelines/augmentations.py
innerlee: Okay, we can deprecate it in next version and remove it in the next next version

dreamerlin: https://github.com/open-mmlab/mmaction2/pull/787

mmaction/datasets/pipelines/augmentations.py
kennymckormick: That's OK. Besides, although these classes are provided, they are never used alone in the pipeline,

mmaction/datasets/pipelines/augmentations.py
kennymckormick: I'll change it here, in this PR.

kennymckormick:
kindly ping @innerlee

innerlee:
Ok ok

kennymckormick:
kindly ping @innerlee

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -42,6 +43,39 @@ def _init_lazy_if_proper(results, lazy):
         assert 'lazy' not in results, 'Use Fuse after lazy operations'
 
 
+@PIPELINES.register_module()
+class EntityBoxRescale:
+
+    def __init__(self, scale_factor):
+        warnings.warn('"EntityBoxRescale" is deprecated and will be removed '
```
innerlee, line 50: Make this an error, since the code will not behave correctly.

innerlee: And provide this pr link in the error message. `"Details see https://github.com/open-mmlab/mmaction2/pull/782"`

innerlee:
Good to go after minor change

mmaction/datasets/pipelines/augmentations.py
kennymckormick: Actually, this module is never used in the data pipeline, and 
**it never behaves correctly in the data pipeline (when used alone)**. 
I think no one has used it in the data pipeline.

innerlee (approved):

PR #800, opened 2021-04-12
[Fix] Fix a bug in pytorch2onnx.py when `num_classes <= 4`

irvingzhang0512:
When `num_classes <= 4`, `pytorch_result[:, 4]` will raise index out of range error.

### DIFF
--- a/tools/pytorch2onnx.py
+++ b/tools/pytorch2onnx.py
@@ -94,8 +94,9 @@ def pytorch2onnx(model,
         onnx_result = sess.run(
             None, {net_feed_input[0]: input_tensor.detach().numpy()})[0]
         # only compare part of results
+        random_class = np.random.randint(pytorch_result.shape[1])
         assert np.allclose(
-            pytorch_result[:, 4], onnx_result[:, 4]
+            pytorch_result[:, random_class], onnx_result[:, random_class]
         ), 'The outputs are different between Pytorch and ONNX'
         print('The numerical values are same between Pytorch and ONNX')
 
--- a/tools/pytorch2onnx.py
+++ b/tools/pytorch2onnx.py
@@ -81,7 +81,7 @@ def pytorch2onnx(model,
 
         # check the numerical value
         # get pytorch output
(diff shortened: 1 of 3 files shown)


tools/pytorch2onnx.py
```diff
@@ -81,7 +81,7 @@ def pytorch2onnx(model,
 
         # check the numerical value
         # get pytorch output
-        pytorch_result = model(input_tensor)[0].detach().numpy()
+        pytorch_result = model(input_tensor).detach().numpy()
```
irvingzhang0512, line 84: For now we need this `[0]` because `dummy_forward` return a tuple.

https://github.com/open-mmlab/mmaction2/blob/c4978a54d0bfdd5a19cb7c64981fcba7350cf78e/mmaction/models/recognizers/recognizer2d.py#L163

But I don't understand why we return a tuple.

SuX97 (approved):

tools/pytorch2onnx.py
SuX97: I have no idea why it is necessary, seems that this `forward_dummy` is only used in flops calculation and model deployment. Any comments? @dreamerlin

innerlee: Since later we may have a universal tool for all repos, keep consistency with other repos and within this repo is important.

dreamerlin: It will be not equal with `onnx_result` if we remove `[0]`. BTW, `model(input_tensor)` will call `forward` rather than `forward_dummy`

irvingzhang0512: @dreamerlin In pytorch2onnx.py, we replace `forward` with `forward_dummy`, so `model(input_tensor)` will call `forward_dummy`

https://github.com/open-mmlab/mmaction2/blob/9ee7129e6ad86131154d00a0e809f082c801441b/tools/pytorch2onnx.py#L149-L152

dreamerlin: Ops, my fault.

dreamerlin (approved):

Issue #806, opened 2021-04-13
FileNotFoundError for AVA dataset


LeeJunHyun:
Thanks for commitments of you all.

I got a FileNotFoundError when processing AVA dataset for LFB inference.

```
python tools/test.py configs/detection/lfb/lfb_slowonly_r50_ava_infer.py \
    checkpoints/slowonly_kinetics_pretrained_r50_4x16x1_20e_ava_rgb_20201217-40061d5f.pth --eval mAP
```

I followed your [guideline](https://github.com/open-mmlab/mmaction2/blob/master/tools/data/ava/README.md) to prepare AVA dataset and it worked for the other frames but not for `data/ava/rawframes/-IELREHX_js/img_22536.jpg`.

I think there are some missing frames but data_loader is not holding the list of valid frames.

For other datasets, there are `*_val_list.txt` and ` *_train_list.txt ` but not for AVA dataset.

Is there any step that I missed?

I attached full error message below.

```
  File "tools/test.py", line 356, in <module>
    main()
  File "tools/test.py", line 341, in main
    outputs = inference_pytorch(args, cfg, distributed, data_loader)
  File "tools/test.py", line 152, in inference_pytorch
    outputs = single_gpu_test(model, data_loader)
  File "/data/private/mmaction2/mmaction/apis/test.py", line 30, in single_gpu_test
    for data in data_loader:
  File "/data/private/mmaction2/mmaction/datasets/data_loader.py", line 345, in __next__
    data = self._next_data()
  File "/data/private/mmaction2/mmaction/datasets/data_loader.py", line 857, in _next_data
    return self._process_data(data)
  File "/data/private/mmaction2/mmaction/datasets/data_loader.py", line 884, in _process_data
    data.reraise()
  File "/opt/conda/envs/open-mmlab/lib/python3.7/site-packages/torch/_utils.py", line 395, in reraise
    raise self.exc_type(msg)
FileNotFoundError: Caught FileNotFoundError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/opt/conda/envs/open-mmlab/lib/python3.7/site-packages/torch/utils/data/_utils/worker.py", line 178, in _worker_loop
    data = fetcher.fetch(index)
  File "/opt/conda/envs/open-mmlab/lib/python3.7/site-packages/torch/utils/data/_utils/fetch.py", line 44, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
[...]

PR #576, opened 2021-01-28
Update cut_videos.sh

kennymckormick:
Remove frame rate adjusting in cut_videos.sh

PR #807, opened 2021-04-14
Fix #534

### LINKED ISSUE #534: Still some bugs during AVA training
I reported some bugs during AVA training last time. Finally(I rename the image files manually), I can run the command 
"./tools/dist_train.sh
configs/detection/ava/slowfast_kinetics_pretrained_r50_8x8x1_20e_ava_rgb.py 4 --validate" in my PC. But during Epoch[1] [120/11524], it raise a bug "FileNotFoundError: [Errno 2] No such file or directory: '/home/avadata/ava/rawframes/7g37N3eoQ9s/img_26368.jpg'". It seems like There are still some bugs in the file name correspondence.  
BTW, I find that in the config file I'm using, line89 and line 90 are as follows :
line89: # Rename is needed to use mmdet detectors
line90:    dict(type='Rename', mapping=dict(imgs='img'))
These codes maybe use for change the file name from ${video_name}_00001.jpg to img_00001.jpg(I guess). But actually, it does not work for some reasons. And I can not find a module named "Rename" in mmdet. 
Hope you can check the questions, thanks a lot.

irvingzhang0512:
It seems that #576 haven't fixed #534 since it doesn't add `-r 30` after `-i`

Cut videos: ffmpeg
Extract frames: ffmpeg or opencv

#576 fix cut with ffmpeg + extract with ffmpeg
According to #806, cut with ffmpeg + extract with opencv couldn't work.
Will have a try

### DIFF
--- a/tools/data/ava/cut_videos.sh
+++ b/tools/data/ava/cut_videos.sh
@@ -29,6 +29,6 @@ for video in $(ls -A1 -U ${IN_DATA_DIR}/*)
 do
   out_name="${OUT_DATA_DIR}/${video##*/}"
   if [ ! -f "${out_name}" ]; then
-    ffmpeg -ss 900 -t 901 -i "${video}" -strict experimental "${out_name}"
+    ffmpeg -ss 900 -t 901 -i "${video}" -r 30 -strict experimental "${out_name}"
   fi
 done


kennymckormick:
I haven't used OpenCV for RGB frame extraction yet. I remember the cons is that OpenCV is slower than FFmpeg, but I'm not sure if it leads to better recognition performance. Have you guys used OpenCV for RGB frame extraction before? @innerlee @dreamerlin @congee524 @SuX97 @irvingzhang0512, what are the pros and cons for OpenCV-based RGB frame extraction?

innerlee:
Either is okay to me. The algorithm should be genralizable to most standard decoder. If it performs good on opencv but poor on ffmpeg, then what's the point of the algorithm? In reality you may read directly from webcam or any other sources, the decoding or video formatting might be much wilder. I think it is not about performance, but about convenience.

irvingzhang0512:
I used opencv to extract frames once and won't use it again... Compared to ffmepg, opencv is slow and inconvenient.

innerlee (approved):

Issue #808, opened 2021-04-14
hmdb51数据集生成帧文件列表时


wangwenhua1:
[Errno 2] No such file or directory: '/hy-nas/mmaction2/data/hmdb51/annotations/trainlist01.txt'

dreamerlin:
Try to remove 'data/hmdb51/annotations/classInd.txt' and try again?

PR #811, opened 2021-04-14
make hmdb train/val list generation more robust

dreamerlin:


kennymckormick:
Perhaps I didn't get it. Why the old codes are less robust and how this PR is related to Issue #808. Can you add some explanation to help me understand it?

dreamerlin:
@kennymckormick If 'data/hmdb51/annotations/classInd.txt' is generated once and `data/hmdb51/annotations/trainlist{:02d}.txt` / `data/hmdb51/annotations/testlist{:02d}.txt` happen to be lost (program crash, file transmisson, etc).  These two lines will not be go through.

https://github.com/open-mmlab/mmaction2/blob/8972e537a77770c4488b33aa595ae45054239159/tools/data/parse_file_list.py#L472-L473

which may cause the failure of generating `data/hmdb51/annotations/trainlist{:02d}.txt` / `data/hmdb51/annotations/testlist{:02d}.txt`

tools/data/parse_file_list.py
```diff
@@ -430,15 +430,23 @@ def generate_class_index_file():
...
+                    cur_line = ' '.join([str(class_id + 1), class_name])
+                    content.append(cur_line)
+                content = '\n'.join(content)
+                f.write(content)
+        else:
+            print(f'{class_index_file} has been generated before.')
+            class_dict = {
+                class_name: class_id + 1
+                for class_id, class_name in enumerate(class_list)
+            }
```
congee524, line 449: we can define class_dict first and once
```python
class_dict = {
    class_name: class_id + 1
    for class_id, class_name in enumerate(class_list)
}
if not osp.exists(class_index_file):
    ...
else:
    ...
```

congee524:
Maybe it's caused by other problems. We can keep the original issue #808 first

congee524 (approved):

kennymckormick:
> @kennymckormick If 'data/hmdb51/annotations/classInd.txt' is generated once and `data/hmdb51/annotations/trainlist{:02d}.txt` / `data/hmdb51/annotations/testlist{:02d}.txt` happen to be lost (program crash, file transmisson, etc). These two lines will not be go through.
> 
> https://github.com/open-mmlab/mmaction2/blob/8972e537a77770c4488b33aa595ae45054239159/tools/data/parse_file_list.py#L472-L473
> 
> which may cause the failure of generating `data/hmdb51/annotations/trainlist{:02d}.txt` / `data/hmdb51/annotations/testlist{:02d}.txt`

got it

kennymckormick (approved):

PR #819, opened 2021-04-19
[Improvement] Add link to TensorRT implementation of TSM-R50

irvingzhang0512:
## Description

For TensorRT 7.2, it's ok to convert TSM-R50 PyTorch -> ONNX -> TensorRT engine. However, I failed to convert TSM-R50 ONNX model to TensorRT engine with TensorRT 5.1.

In order to overcome this issue, I'm trying to build TSM-R50 with TensorRT network definition API. For now, all we need to do is convert pytorch checkpoints to TensorRT weights and don't have to convert PyTorch -> ONNX -> TensorRT

### DIFF
--- a/configs/recognition/tsm/README.md
+++ b/configs/recognition/tsm/README.md
@@ -142,3 +142,5 @@ python tools/test.py configs/recognition/tsm/tsm_r50_1x1x8_50e_kinetics400_rgb.p
 ```
 
 For more details, you can refer to **Test a dataset** part in [getting_started](/docs/getting_started.md#test-a-dataset).
+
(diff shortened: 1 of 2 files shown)


innerlee:
Thanks for providing the info. This is good for a post in [discussions](https://github.com/open-mmlab/mmaction2/discussions). 
Methods' README is used to give information about the supported configs, so I'm hesitate to add this since this thrid-party repo is not tested/endorsed/maintained by us.

innerlee:
We will provide utilities for tensorrt support in future, and the current design of that repo is not compatible to ours.

PR #820, opened 2021-04-19
[Fix] Fix bug: Resume can not work.

kennymckormick:
The first line of cfg.text is the config filename, which will not be parsed by this line:
https://github.com/open-mmlab/mmcv/blob/de4f14e9cd547fb8bf886a362ad4f02a0be95141/mmcv/runner/base_runner.py#L346

Use cfg.pretty_text instead.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -137,7 +137,7 @@ def main():
 
     # log some basic info
     logger.info(f'Distributed training: {distributed}')
-    logger.info(f'Config: {cfg.text}')
+    logger.info(f'Config: {cfg.pretty_text}')
 
     # set random seeds
     if args.seed is not None:


innerlee (approved):

sssste:
你好，我修改成pretty_text以后还是不能resume, 还是识别不了第一行

kennymckormick:
> 你好，我修改成pretty_text以后还是不能resume, 还是识别不了第一行

My bad, I didn't replace it completely, you also need to replace `text` in `cfg.checkpoint_config.meta` by `pretty_text`. I have raised a patch (#824) for it. 
Besides, you can resume from a ckpt only if the ckpt is trained with pretty_text rather than text.

PR #833, opened 2021-04-25
use twice sample for tpn_tsm sthv1 testing

dreamerlin:


congee524 (approved):

congee524:
Hey, it seems that we didn't do color jitter in test_pipeline while doing it in train & val pipeline.

Is that a problem?

PR #837, opened 2021-04-26
[Docs] fix typo for sthv1 data preparing

LiUzHiAn:
## Typo

Hi, I think the data preprocessing [step-4](https://github.com/open-mmlab/mmaction2/blob/master/tools/data/sthv1/README.md#L87-L97) for something-v1 is for video format input，so I fix the typos accordingly.

### DIFF
--- a/tools/data/sthv1/README.md
+++ b/tools/data/sthv1/README.md
@@ -86,7 +86,7 @@ bash extract_flow.sh
 
 ## Step 4. Encode Videos
 
-This part is **optional** if you only want to use RGB frames.
+This part is **optional** if you only want to use RGB videos.
 
 You can run the following script to encode videos.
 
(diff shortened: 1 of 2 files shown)


dreamerlin:
The original description is correct

tools/data/sthv1/README_zh-CN.md
```diff
@@ -61,7 +61,7 @@ data = dict(
 
 ## 步骤 3. 抽取光流
 
-如果用户只想使用原 RGB 帧加载训练，则该部分是 **可选项**。
+如果用户只想使用原 RGB 视频加载训练，则该部分是 **可选项**。
```
innerlee, line 64: This is step3.

tools/data/sthv1/README.md
```diff
@@ -86,7 +86,7 @@ bash extract_flow.sh
 
 ## Step 4. Encode Videos
 
-This part is **optional** if you only want to use RGB frames.
+This part is **optional** if you only want to use RGB videos.
```
innerlee, line 89: It is correct, since this step will produce rgb videos. It is optional if you only uses frames.

PR #856, opened 2021-05-07
[Feature] Feature extraction

kennymckormick:
Here we provide an example of how to perform feature extraction with pre-trained video recognizers.


tools/clip_feature_extraction.py
```diff
@@ -0,0 +1,223 @@
...
+
+from mmaction.apis import multi_gpu_test, single_gpu_test
+from mmaction.datasets import build_dataloader, build_dataset
+from mmaction.models import build_model
+from mmaction.utils import register_module_hooks
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='MMAction2 test (and eval) a model')
```
congee524, line 26: Modify the description.

mmaction/models/recognizers/recognizer2d.py
```diff
@@ -70,20 +72,31 @@ def _do_test(self, imgs):
             x = x.squeeze(2)
             num_segs = 1
 
+        if self.feature_extraction:
+            # perform spatial pooling
+            avg_pool = nn.AdaptiveAvgPool2d(1)
```
congee524, line 77: _do_test will be called multi times in recognize2d & 3d, maybe we can move avg_pool to __init__

mmaction/models/recognizers/recognizer2d.py
```diff
@@ -70,20 +72,31 @@ def _do_test(self, imgs):
...
+            x = avg_pool(x)
+            # squeeze dimensions
+            x = x.reshape((batches, num_segs, -1))
+            # temporal average pooling
+            x = x.mean(axis=1)
+            return x
```
congee524, line 83: To make users more flexible to handle these features, maybe we can return x directly and save them without avg_pool reshape mean. Is there some flexible method?

mmaction/models/recognizers/recognizer2d.py
```diff
@@ -108,6 +121,8 @@ def _do_fcn_test(self, imgs):
             x = x.reshape((-1, num_segs) +
                           x.shape[1:]).transpose(1, 2).contiguous()
 
+        assert not self.feature_extraction
+        assert self.with_cls_head
```
congee524, line 125: place the two lines in `forward_test` to terminate the program early.

kennymckormick:
kindly ping @innerlee

innerlee:
Some general observations/questions?

* Is it possible to implement a general feature extraction tool, without the need to write a new config?
* There are some avg ops hard-coded. Can they simply respect the settings in the original config?
* There was an implementation of returning features in mmpose https://github.com/open-mmlab/mmpose/pull/212 I'm not sure if it can help, or give some hints

kennymckormick:
> Some general observations/questions
> 
> * Is it possible to implement a general feature extraction tool, without the need to write a new config?
> * There are some avg ops hard-coded. Can they simply respect the settings in the original config?
> * There was an implementation of returning features in mmpose [open-mmlab/mmpose#212](https://github.com/open-mmlab/mmpose/pull/212) I'm not sure if it can help, or give some hints

1. It is possible to do so. However, I think writing a separate config for feature extraction can be clearer (like a separate config for inference). Besides, we only need one config for a backbone, and it can deal with weights pretrained on any datasets.
2. For now, there are some hard-coded avg ops in the code, since our feature extraction will return an n-dim vector for each clip by default (which is most often used).  In the future we may add more options for feature extraction.
3. A little bit different here: in mmpose, users usually want both keypoint predictions and heatmaps/features. However, when you extract video features, usually you just want the feature and the predicted class might be meaningless.

innerlee:
I'm not a fan of writing many configs. My philosophy is that, for peripheral codes, they should "just-work". The "inference" example is a bad example. The design of inference is not good if it need a separate config.

That said, this can be get merged by @dreamerlin or @hellock .

kennymckormick:
> I'm not a fan of writing many configs. My philosophy is that, for peripheral codes, they should "just-work". The "inference" example is a bad example. The design of inference is not good if it need a separate config.
> 
> That said, this can be get merged by @dreamerlin or @hellock .

So what about make `feature_extraction` an optional argument. Then we will not need a separate config file.

innerlee:
> ... will not need a separate config file.

This will be great.

kennymckormick:
@innerlee. Now there exists a flag named 'feature_extraction' in model arguments. The script set it to True and it can handle any config files. However, I still keep the two additional config files as examples of **minimal** config files for feature extraction.

mmaction/models/recognizers/recognizer2d.py
```diff
@@ -11,6 +11,8 @@ class Recognizer2D(BaseRecognizer):
 
     def forward_train(self, imgs, labels, **kwargs):
```
dreamerlin, line 12: Is it right to put `feature_extraction` to `forward_train` rather than `forward_test` ?

mmaction/models/recognizers/base.py
```diff
@@ -22,20 +22,26 @@ class BaseRecognizer(nn.Module, metaclass=ABCMeta):
...
-                 cls_head,
+                 cls_head=None,
+                 feature_extraction=False,
                  neck=None,
                  train_cfg=None,
                  test_cfg=None):
         super().__init__()
         # record the source of the backbone
         self.backbone_from = 'mmaction2'
+        self.feature_extraction = feature_extraction
```
dreamerlin, line 44: How about put `feature_extraction` in `test_cfg`?

mmaction/models/recognizers/recognizer2d.py
dreamerlin: How about `F.adaptive_avg_pool2d`

mmaction/models/recognizers/recognizer3d.py
```diff
@@ -38,22 +41,46 @@ def _do_test(self, imgs):
...
+            feat = self.extract_feat(imgs)
             if self.with_neck:
-                x, _ = self.neck(x)
-            cls_score = self.cls_head(x)
-
+                feat, _ = self.neck(feat)
+
+        if self.feature_extraction:
+            # perform spatio-temporal pooling
+            avg_pool = nn.AdaptiveAvgPool3d(1)
```
dreamerlin, line 68: F.adaptive_avg_pool3d

mmaction/models/recognizers/base.py
kennymckormick: Good Idea.

mmaction/models/recognizers/recognizer3d.py
kennymckormick: I guess there is little performance difference 😄

Issue #862, opened 2021-05-11
Output predictions as json (for each frame)


rlleshi:
The supported outputs currently are  _video_ or _gif_ formats. Could we in addition have as an output _json_ files? 

Using `long_video_demo` for example the output can also be a  _json_ file that contains predictions for every frame of the input video.

kennymckormick:
Good suggestion, we will consider adding this feature. Besides, if you have already implemented this, you are welcome to raise a PR.

rlleshi:
sure

PR #906, opened 2021-06-05
862 add json output video demo

rlleshi:
## Motivation
#862 
Currently `long_video_demo.py` only outputs videos. But it will also be beneficial to have it output a JSON file with predictions for each frame.

## Modification

Modified `long_video_demo.py` by adding a new function to keep track of predictions and write them to a JSON file.

## Use cases (Optional)

Call `long_video_demo.py ` by providing a `.json` file as `OUT_FILE`

### DIFF
--- a/demo/long_video_demo.py
+++ b/demo/long_video_demo.py
@@ -1,4 +1,5 @@
 import argparse
+import json
 import random
 from collections import deque
 from operator import itemgetter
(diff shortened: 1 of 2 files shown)


dreamerlin:
Thanks for this PR. Could you please also provide an example command in the document ?

dreamerlin (commented):
in line# 120, `video_writer` should be None when `args.out_file` is not a Video

demo/long_video_demo.py
```diff
@@ -70,6 +112,7 @@ def show_results(model, data, label, args):
 
     msg = 'Preparing action recognition ...'
     text_info = {}
+    out_json = {}
     fourcc = cv2.VideoWriter_fourcc(*'mp4v')
     frame_size = (frame_width, frame_height)
```
dreamerlin, line 118: in line#120, `video_writer` should be `None` when `args.out_file` is not a Video

dreamerlin (approved):

dreamerlin:
Thanks

PR #908, opened 2021-06-07
[Improvement] Use Pylint to polish code style

dreamerlin:


mmaction/core/evaluation/eval_hooks.py
```diff
@@ -40,6 +39,8 @@ def __init__(self, *args, save_best='auto', **kwargs):
 
 if not from_mmcv:
 
+    from mmcv.runner import Hook
```
congee524, line 42: There should be many imported packages that are only used in local code blocks. Do you want to modify them all?

dreamerlin: Some of them may cause circle import, so just keep it.

mmaction/datasets/audio_visual_dataset.py
```diff
@@ -65,7 +65,7 @@ def load_annotations(self):
                     idx += 1
                 # idx for label[s]
                 label = [int(x) for x in line_split[idx:]]
-                assert len(label), f'missing label in line: {line}'
+                assert len(label) != 0, f'missing label in line: {line}'
```
congee524, line 68: maybe `> 0` is better, but it's not a big problem

dreamerlin: just format according to Pylint which is based on PEP8

kennymckormick:
Is this PR still in WIP? 
image

kennymckormick (approved):

Issue #904, opened 2021-06-04
inference_recognizer on optical flow model is throwing  "FileNotFoundError, No such file or directory"  accessing non existent frame on rawframe video


olguin:
I finetuned a model pretrained on your slowonly_r50_8x8x1_256e_kinetics400_flow  with 2 custom classes. Calculated optical flow using build_rawframe script for input videos, Training, validation and test on test dataset all worked very well. 

I got an error when I tried  to use the pretrained model to make a prediction on a single video (which I had previously calculated its optical flow flow_x_%5d.txt and flow_y_%5d.txt  files using build_rawframes.py script). 

Here is the code for the prediction I am trying to get:

```from mmaction.apis import init_recognizer, inference_recognizer
import torch
device = 'cuda:0' # or 'cpu'
device = torch.device(device)

config_file = 'mmaction2/configs/recognition/slowonly/screw_slowonly_r50_8x8x1_256e_kinetics400_flow.py'
checkpoint_file = 'mmaction2/working_dir/slowonly_screwdriver_flow_run4/best_top1_acc_epoch_75.pth'
model = init_recognizer(config_file, checkpoint_file, device=device, use_frames=True)
video = '/home/ubuntu/dataset/screw9/rawframes/L4IloTT-5zs_1028_1030'
label = '/home/ubuntu/dataset/screw/screw_label_map.txt'
results = inference_recognizer(model, video, label, use_frames=True)
for result in results:
    print(f'{result[0]}: ', result[1])
```

When running that I am getting the error:
**Error traceback**
```
No such file or directory: '/home/ubuntu/dataset/screw/rawframes/L4IloTT-5zs_1028_1030/flow_x_00059.jpg'
```

The things is that  the video L4IloTT-5zs_1028_1030 has only 57 frames (and the frames go from flow_x_00000 to flow_x_00056). 

I think that the error is in the lines:
```~/mmaction2/mmaction/datasets/pipelines/loading.py in __call__(self, results)
   1146                     y_filepath = osp.join(directory,
   1147                                       filename_tmpl.format('y', frame_idx))
-> 1148                     x_img_bytes = self.file_client.get(x_filepath)
```

If I print the list of frame_idx for this video I am getting :

```frames [  3  11  19  27  35  43  51  59   8  16  24  32  40  48  56  64  13  21
  29  37  45  53  61  69  18  26  34  42  50  58  66  74  23  31  39  47
[...]

PR #922, opened 2021-06-11
Support number counting for flow-wise filename template

irvingzhang0512:
## Motivation

1. As #904 and the following comment shows, for now, in `inference_recognizer` we assumes files in `video_path` are all named with filename_tmpl. This pr removes the above restriction.

https://github.com/open-mmlab/mmaction2/blob/f007661ab97b206bfb82ed7f815999e2d7964a9b/mmaction/apis/inference.py#L112

2. We have already provided a way to fix bug in #565. However, we haven't modified related configs. So if we train flow models with default config, this shape mismatch bug still occurs.

## Modification

1. For `inference_recognizer`, this pr count the file that filename format matches `filename_impl`, add some unittests. 
2. For TSN Flow configs, this pr modifies `in_channels` of `backbone`


dreamerlin (approved):

dreamerlin:
Thanks for your contribution for MMAction2!

PR #935, opened 2021-06-17
[Improvement] Adjust script structure

dreamerlin:
Add directories for scripts for `tools`, only leaving `train` and `test` related files


configs/localization/bsn/README.md
```diff
@@ -71,7 +71,7 @@ You can use the following commands to inference a model.
 2. For PGM Inference
 
     ```shell
-    python tools/bsn_proposal_generation.py ${CONFIG_FILE} [--mode ${MODE}]
+    python tools/misc/bsn_proposal_generation.py ${CONFIG_FILE} [--mode ${MODE}]
```
innerlee, line 74: misc is not a good name

dreamerlin: Any good idea?

dreamerlin: Let's keep it as `misc` ?

kennymckormick: I think misc is OK.

kennymckormick (commented):
Suggest not to place `feature_extraction` in  `misc`, it's an important feature like training or testing. Besides, the `dist_clip_feature_extraction.sh`  will not work now since the PYTHONPATH should also be altered.

kennymckormick (approved):

PR #939, opened 2021-06-17
[Improvement] Change ResNet3D default value for stem with BC

dreamerlin:


dreamerlin:
@kennymckormick need double-checking

innerlee:
Any reason?

dreamerlin:
Our ResNet3D refers to https://github.com/open-mmlab/mmaction/blob/master/mmaction/models/tenons/backbones/resnet_i3d.py, whose stride is 2. And Since we use ResNet3D as a basic module, it is more common to set those as 1 for usage. @kennymckormick What do you think of?

kennymckormick (approved):

kennymckormick:
> Our ResNet3D refers to https://github.com/open-mmlab/mmaction/blob/master/mmaction/models/tenons/backbones/resnet_i3d.py, whose stride is 2. And Since we use ResNet3D as a basic module, it is more common to set those as 1 for usage. @kennymckormick What do you think of?

I think changing the default value is better.

innerlee:
Old ckpts will suddenly refuse to work, no?

dreamerlin:
> Old ckpts will suddenly refuse to work, no?

It doesn't affect the old ckpts, since current configs in the PRs are the same with the old ones.

irvingzhang0512:
FYI SlowFast I3D's stem maxpooling uses `kernel_size=[1, 3, 3], stride=[1, 2, 2], padding=[0, 1, 1]`.
Related codes are [here](https://github.com/facebookresearch/SlowFast/blob/923e9470b30ac7749af563362eed77339375c82e/slowfast/models/stem_helper.py#L192-L194)

PR #943, opened 2021-06-17
Audit the usage of shutil.rmtree(tmp_dir)

dreamerlin:


mmaction/apis/test.py
```diff
@@ -131,6 +131,7 @@ def collect_results_cpu(result_part, size, tmpdir=None):  # noqa: F811
             dist.broadcast(dir_tensor, 0)
             tmpdir = dir_tensor.cpu().numpy().tobytes().decode().rstrip()
         else:
+            tmpdir = osp.join('.dist_test', tmpdir)
```
innerlee, line 134: Wrong order of join?

tools/data/hvu/download.py
```diff
@@ -149,7 +149,10 @@ def main(input_csv,
          output_dir,
          trim_format='%06d',
          num_jobs=24,
-         tmp_dir='/tmp/hvu'):
+         tmp_dir='hvu'):
+
+    tmp_dir = os.path.join('/tmp', tmp_dir)
```
kennymckormick, line 154: We should not enforce the tmp_dir to be under the `/tmp/` directory: 
1. What if the user wants to put it in another place, i.e., because little space left in `/`?
2. Poor compatibility: if we want MMAction2 to run on Windows in the future, we need to change here again.

tools/data/kinetics/download.py
```diff
@@ -175,7 +175,9 @@ def main(input_csv,
          output_dir,
          trim_format='%06d',
          num_jobs=24,
-         tmp_dir='/tmp/kinetics'):
+         tmp_dir='kinetics'):
+    tmp_dir = os.path.join('/tmp', tmp_dir)
```
kennymckormick, line 179: Same as above.

mmaction/apis/test.py
dreamerlin: correct it to `osp.join(tmpdir, '.dist_test')`

tools/data/hvu/download.py
dreamerlin: `/tmp` -> `./tmp`

dreamerlin: `tmp_dir = os.path.join(tmp_dir, '.tmp_dir')` now

dreamerlin:
@innerlee @kennymckormick can this be merged now?

kennymckormick (approved):

PR #945, opened 2021-06-18
[ModelZoo] Port CSN checkpoint from VMZ

dreamerlin:


dreamerlin:
These files are directly ported from VMZ without training, are they still needed to write into meta files?

dreamerlin:
@kennymckormick

kennymckormick:
Please remove the fields like gpu_mem and inference_time for now (since they are not systematically evaluated now). Also, the field 'FLOPs' should be included.

kennymckormick (approved):

PR #947, opened 2021-06-20
[Improvement] Add font color to args in long_video_demo

rlleshi:
## Motivation

Make font color in `long_video_demo.py` controllable by fetching it as args to the script. Otherwise one has to modify the code.

## Modification

Small modifications at `long_video_demo.py`

## Use cases (Optional)

Updated docs accordingly

### DIFF
--- a/demo/long_video_demo.py
+++ b/demo/long_video_demo.py
@@ -68,12 +68,24 @@ def parse_args():
         type=int,
         default=(255, 255, 255),
         help='font color (B, G, R) of the labels in output video')
+    parser.add_argument(
+        '--msg-color',
+        nargs='+',
+        type=int,
+        default=(128, 128, 128),
+        help='font color (B, G, R) of the messages in output video')
     args = parser.parse_args()
     return args
 
 
-def show_results_video(result_queue, text_info, thr, msg, frame, video_writer,
-                       font_color):
+def show_results_video(result_queue,
+                       text_info,
+                       thr,
+                       msg,
+                       frame,
+                       video_writer,
+                       font_color=(255, 255, 255),
+                       msg_color=(128, 128, 128)):
     if len(result_queue) != 0:
         text_info = {}
         results = result_queue.popleft()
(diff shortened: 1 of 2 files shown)


demo/long_video_demo.py
```diff
@@ -79,20 +83,20 @@ def show_results_video(result_queue, text_info, thr, msg, frame, video_writer):
...
-                    THICKNESS, LINETYPE)
-    video_writer.write(frame)
+        cv2.putText(fr, msg, (0, 40), FONTFACE, FONTSCALE, clr, THICKNESS,
+                    LINETYPE)
+    v_writer.write(fr)
     return text_info
 
 
-def get_results_json(result_queue, text_info, thr, msg, ind, out_json):
+def get_results_json(result_queue, text_info, msg, thr, ind, out_json):
```
dreamerlin, line 99: It is not good to change the argument position

demo/long_video_demo.py
```diff
@@ -79,20 +83,20 @@ def show_results_video(result_queue, text_info, thr, msg, frame, video_writer):
             location = (0, 40 + i * 20)
             text = selected_label + ': ' + str(round(score, 2))
             text_info[location] = text
-            cv2.putText(frame, text, location, FONTFACE, FONTSCALE, FONTCOLOR,
+            cv2.putText(fr, text, location, FONTFACE, FONTSCALE, clr,
                         THICKNESS, LINETYPE)
     elif len(text_info):
         for location, text in text_info.items():
-            cv2.putText(frame, text, location, FONTFACE, FONTSCALE, FONTCOLOR,
+            cv2.putText(fr, text, location, FONTFACE, FONTSCALE, clr,
```
dreamerlin, line 90: it is not good to use abbreviation as variable name, which is hard to understand

demo/long_video_demo.py
rlleshi: I updated accordingly. But why isn't this one good btw?

demo/long_video_demo.py
```diff
@@ -64,11 +62,18 @@ def parse_args():
...
+        type=int,
+        default=(255, 255, 255),
+        help='font color (B, G, R) of the labels in output video')
     args = parser.parse_args()
     return args
 
 
-def show_results_video(result_queue, text_info, thr, msg, frame, video_writer):
+def show_results_video(result_queue, text_info, thr, msg, font_color, frame,
+                       video_writer):
```
dreamerlin, line 76:
```python
def show_results_video(result_queue, text_info, thr, msg, frame, video_writer, font_color=(255, 255, 255)):
```

rlleshi: ah.. updated (default value is in `parse_args`, however)

innerlee:
I'm okay with the change as long as it can run successfully 😃

PR #950, opened 2021-06-21
Improve check videos

rlleshi:
There are some problems with the `check_videos.py` script. I didn't see any current PR or issues on it so made this one.
Also improved some minor stuff.

## Bugs
- [pool.join()](https://github.com/open-mmlab/mmaction2/blob/master/tools/check_videos.py#L143) was being called before [pool.close()](https://github.com/open-mmlab/mmaction2/blob/master/tools/check_videos.py#L146). It should be the other way around. This also leads to the following error:

```
  File "tools/check_videos.py", line 143, in <module>
    pool.join()
  File "/opt/conda/lib/python3.8/multiprocessing/pool.py", line 659, in join
    raise ValueError("Pool is still running")
ValueError: Pool is still running
```

- Getting file length
The current way, [len(f)](https://github.com/open-mmlab/mmaction2/blob/master/tools/check_videos.py#L149) doesn't work since objects don't have length. Corresponding error:

```
  File "tools/check_videos.py", line 148, in <module>
    print(f'Checked {len(dataset)} videos, '
TypeError: object of type '_io.TextIOWrapper' has no len()

```
Added a function to get file length.

- If no clips with problems were found then the script fails because no output file exists, e.g. [here](https://github.com/open-mmlab/mmaction2/blob/master/tools/check_videos.py#L147).
- Same thing as above for deleting corrupted clips when no output file exists

## Minor improvements
- Remove [num-processes](https://github.com/open-mmlab/mmaction2/blob/master/tools/check_videos.py#L46) as args and add instead all but one cpus for multiprocessing: `pool = Pool(cpu_count() - 1 or 1)`
- Add choices for args [here ](https://github.com/open-mmlab/mmaction2/blob/master/tools/check_videos.py#L42) and [here](https://github.com/open-mmlab/mmaction2/blob/master/tools/check_videos.py#L38)

### DIFF
--- a/demo/long_video_demo.py
+++ b/demo/long_video_demo.py
@@ -109,7 +109,7 @@ def show_results_video(result_queue,
     return text_info
 
 
-def get_results_json(result_queue, text_info, msg, thr, ind, out_json):
+def get_results_json(result_queue, text_info, thr, msg, ind, out_json):
     if len(result_queue) != 0:
         text_info = {}
         results = result_queue.popleft()
@@ -180,7 +180,7 @@ def show_results(model, data, label, args):
 
         if args.out_file.endswith('.json'):
             text_info, out_json = get_results_json(result_queue, text_info,
-                                                   msg, args.threshold, ind,
+                                                   args.threshold, msg, ind,
                                                    out_json)
         else:
             text_info = show_results_video(result_queue, text_info,
--- a/demo/long_video_demo.py
+++ b/demo/long_video_demo.py
@@ -109,7 +109,7 @@ def show_results_video(result_queue,
     return text_info
 
 
-def get_results_json(result_queue, text_info, thr, msg, ind, out_json):
+def get_results_json(result_queue, text_info, msg, thr, ind, out_json):
     if len(result_queue) != 0:
         text_info = {}
         results = result_queue.popleft()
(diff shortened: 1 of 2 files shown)


tools/check_videos.py
```diff
@@ -132,28 +135,28 @@ def _do_check_videos(lock, dataset, output_file, idx):
...
+            print('Start deleting corrupted videos')
+            cnt = 0
+            with open(args.output_file, 'r') as f:
+                for line in f:
+                    if os.path.exists(line.strip()):
+                        os.remove(line.strip())
+                        cnt += 1
+            print(f'Deleted {cnt} corrupted videos.')
+    else:
+        print(f'Checked {len(dataset)} videos, ' 'none are corrupted/missing')
```
irvingzhang0512, line 162: `print(f'Checked {len(dataset)} videos, none are corrupted/missing')`

tools/check_videos.py
```diff
@@ -132,28 +135,28 @@ def _do_check_videos(lock, dataset, output_file, idx):
     if os.path.exists(args.output_file):
         # remove exsiting output file
         os.remove(args.output_file)
-    pool = Pool(args.num_processes)
+    pool = Pool(cpu_count() - 1 or 1)
```
irvingzhang0512, line 138: Maybe we can keep `--num-processes` and set default value of `--num-processes` to `cpu_count() - 1`. Sometimes we cannot use all cpu cores.

dreamerlin:
LGTM

dreamerlin (approved):

PR #952, opened 2021-06-22
[Improvement] Make demo more robust in cross-platforms

kennymckormick:
Original demo needs imagemagick to be properly installed (which does not hold on CentOS or wsl-Ubuntu).
This PR use `cv2.putText` instead of `TextClip`


demo/demo.py
```diff
@@ -99,32 +103,45 @@ def get_output(video_path,
...
+        frames = [x.asnumpy()[..., ::-1] for x in video]
+
+    if target_resolution:
+        w, h = target_resolution
+        frame_h, frame_w, _ = frames[0].shape
+        if w == -1:
+            w = int(h / frame_h * frame_w)
+        if h == -1:
+            h = int(w / frame_w * frame_h)
+        frames = [cv2.imresize(f, (w, h)) for f in frames]
```
dreamerlin, line 126: `cv2.imresize` -> `cv2.resize`

demo/demo.py
```diff
@@ -2,7 +2,11 @@
 import os
 import os.path as osp
 
+import cv2
+import decord
+import numpy as np
 import torch
+import webcolors
```
dreamerlin, line 9: `webcolors` should be put in requirements/optional.txt

demo/demo.py
```diff
@@ -65,7 +69,7 @@ def get_output(video_path,
                out_filename,
                label,
                fps=30,
-               font_size=20,
+               font_scale=0.5,
```
dreamerlin, line 72: Need to change related docs

demo/demo.py
dreamerlin: And how about put `decord` and `webcolor` into `try...except` to check whether users have installed them

dreamerlin (changes requested):

demo/demo.py
kennymckormick: webcolors is a very small package (10KB) that map color names to RGB values. Decord is used frequently in mmaction2. Besides, it's easy to install both of them via pip. So that I think we can directly add them to setup.cfg instead

dreamerlin (approved):

dreamerlin:
LGTM and it runs successfully when I test it

PR #966, opened 2021-06-27
[Improvement] Add FLOPs

kennymckormick:
## Motivation

Imporve Model Metadata. 

## Modification

1. Add field FLOPs to recognition models (except tsn_timm)
2. Fix TSM_Jester metafile


dreamerlin:
Is there any script to calculate FLOPs at one time?

kennymckormick:
> Is there any script to calculate FLOPs at one time?

It can be inference with the [MMCV API](https://github.com/open-mmlab/mmcv/blob/master/mmcv/cnn/utils/flops_counter.py), you just need to build the model and specify the input shape.

gengenkai (approved):

PR #972, opened 2021-06-29
[Improvement] Support Torchvision transformations

kennymckormick:
## Motivation

Use torchvision transformations directly in mmaction2.

## Modification

Add a new pipeline component: TorchvisionTrans.
Modify Compose code.
Remove 3 deprecated components: EntityBoxCrop, EntityBoxFlip, EntityBoxRescale

## Drawback

This feature is just used for prototype verification (calling torchvision transformations can be much slower in the preprocessing pipeline).

### DIFF
--- a/mmaction/datasets/pipelines/augmentations.py
+++ b/mmaction/datasets/pipelines/augmentations.py
@@ -1,12 +1,14 @@
 import random
 import warnings
 from collections.abc import Sequence
+from distutils.version import LooseVersion
 
 import mmcv
 import numpy as np
 from torch.nn.modules.utils import _pair
 
 from ..builder import PIPELINES
+from .formating import to_tensor
 
 
 def _combine_quadruple(a, b):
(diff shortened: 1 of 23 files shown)


tests/test_data/test_compose.py
```diff
@@ -4,6 +4,15 @@
...
 from mmaction.datasets.pipelines import Compose, ImageToTensor
 
+try:
+    import torchvision
+    from distutils.version import LooseVersion
+    torchvision_ok = False
+    if LooseVersion(torchvision.__version__) >= LooseVersion('0.8.0'):
+        torchvision_ok = True
+except (ImportError, ModuleNotFoundError):
+    torchvision_ok = False
```
dreamerlin, line 14: No need to try-except

dreamerlin: Alternatives

```python
TORCHVISION_VERSION = tuple(int(x) for x in torchvision.__version__.split('.')[:3])

def obsolete_torchvision_version(torchvision_version, version_threshold):
    return torchvision_version<= version_threshold
```

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -51,6 +53,43 @@ def _init_lazy_if_proper(results, lazy):
         assert 'lazy' not in results, 'Use Fuse after lazy operations'
 
 
+@PIPELINES.register_module()
+class TorchvisionTrans:
+    """Torchvision Augmentations, under torchvision.transforms.
+
+    Args:
+        type: The name of the torchvision transformation.
```
dreamerlin, line 61: type(str)

tests/test_data/test_compose.py
kennymckormick: try except is needed since torchvision should not be added to known-third-parties

dreamerlin (approved):

PR #996, opened 2021-07-06
[Improvement] Remove Flips

kennymckormick:


kennymckormick:
@dreamerlin  Plz take a look.

innerlee:
flop flip

innerlee:
One thing to check is that the modified configs could run successfully, especially for those detection related.
The format bundle in mmdet requires keys like flip directions, so det will raise an error if such keys do not exist.

dreamerlin (approved):

kennymckormick:
> One thing to check is that the modified configs could run successfully, especially for those detection related.
> The format bundle in mmdet requires keys like flip directions, so det will raise an error if such keys do not exist.

Keys added by Flip ('flip', 'flip_direction') is not further used, besides, we haven't changed the detection config files.

PR #1025, opened 2021-07-14
Update colorjitter 

kennymckormick:


mmaction/datasets/pipelines/augmentations.py
```diff
@@ -1448,183 +1449,84 @@ def __repr__(self):
...
-        Returns:
-            np.ndarray: A brightness distorted image.
-        """
-        if np.random.rand() > 0.5:
-            img = img + np.float32(delta)
-        return img
+    def check_input(val, max, base):
+        if isinstance(val, tuple):
+            assert ((val[0] <= val[1]) and (val[0] >= base - max)
+                    and (val[1] <= base + max))
```
dreamerlin, line 1459: `assert base - max <= val[0] <= val[1] <= base + max`

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -1448,183 +1449,84 @@ def __repr__(self):
...
+
+        new_imgs = []
+        for i in range(num_clips):
+            b = np.random.uniform(
+                low=self.brightness[0], high=self.brightness[1])
+            c = np.random.uniform(low=self.contrast[0], high=self.contrast[1])
+            s = np.random.uniform(
+                low=self.saturation[0], high=self.saturation[1])
+            h = np.random.uniform(low=self.hue[0], high=self.hue[1])
+            st, ed = i * clip_len, (i + 1) * clip_len
```
dreamerlin, line 1506: st->start, ed->end

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -1448,183 +1449,84 @@ def __repr__(self):
...
-            self.eig_vec = np.array([[-0.5675, 0.7192, 0.4009],
-                                     [-0.5808, -0.0045, -0.8140],
-                                     [-0.5836, -0.6948, 0.4203]],
-                                    dtype=np.float32)
-        else:
-            self.eig_vec = eig_vec
-
-        self.alpha_std = alpha_std
-        self.color_space_aug = color_space_aug
+    """Perform ColorJitter to each img."""
```
dreamerlin, line 1452: Need to polish the docstring

mmaction/datasets/pipelines/augmentations.py
```diff
@@ -1448,183 +1449,98 @@ def __repr__(self):
...
-            np.dot(self.eig_vec * alpha, self.eig_val), dtype=np.float32)
-        rgb = rgb[None, None, ...]
-
-        results['imgs'] = [img + rgb for img in out]
-        results['eig_val'] = self.eig_val
-        results['eig_vec'] = self.eig_vec
-        results['alpha_std'] = self.alpha_std
-        results['color_space_aug'] = self.color_space_aug
-
+        num_clips, clip_len = 1, len(imgs)
```
dreamerlin, line 1510: `num_clips` is hardcoded to 1 ?

kennymckormick: This `num_clips` is only used in the ColorJitter. Here it indicates we perform the same augmentation for all frames (the same brightness, contrast, saturation, hue factor). If set to the real num_clips, it indicates we perform different augmentations for different clips (in case we sample multiple clips from a video).

kennymckormick: We don't sample multiple clips in training for recognition. However, such operation is needed in self-supervised pretraining, which may be added in the future.

dreamerlin (approved):

dreamerlin:
Better to show an image colorjittered by this augmentation

PR #1026, opened 2021-07-14
Resolve E501 in tools/data/resize_videos.py

makecent:


makecent:
Close because has been solved

PR #1047, opened 2021-07-21
[Enhancement] Calculate Top K accurate and inaccurate classes

dreamerlin:


mmaction/core/evaluation/accuracy.py
```diff
@@ -87,6 +87,64 @@ def mean_class_accuracy(scores, labels):
     return mean_class_acc
 
 
+def top_k_accurate_classes(scores, labels, k=10):
```
kennymckormick, line 90: Since the most part of these two APIs are the same, what about let these two APIs be a wrapper of another API like:
```
top_k_classes(scores, labels, k=10, mode='accurate')
```

kennymckormick (approved):

PR #1073, opened 2021-08-05
[Doc] fix doc

kennymckormick:


docs/stat.py
```diff
@@ -40,7 +40,7 @@ def anchor(name):
         print(p)
         q = p.replace('\\', '\\\\').replace('?', '\\?')
         paperlinks[p] = ' '.join(
-            (f'[⇨]({splitext(basename(f))[0]}.html#{anchor(paperlink)})'
+            (f'[->]({splitext(basename(f))[0]}.html#{anchor(paperlink)})'
```
innerlee, line 43: anything wrong with unicode?

kennymckormick: This char can not be exported to pdf.

dreamerlin:
LGTM as long as it runs successfully.

dreamerlin (approved):

PR #1138, opened 2021-09-08
[Fix] Fix performance number of some checkpoints

kennymckormick:


kennymckormick:
@dreamerlin Hi, in this PR I fix the performance numbers of some checkpoints. Since some of them are provided by you, would you please double-check these updates?

PR #1150, opened 2021-09-14
Fix resume when building rawframes

dreamerlin:


kennymckormick:
Actually, you can not guarantee to resume successfully with this PR. What if the job terminated while extracting a video?  The frame folder is created but the extraction hasn't finished successfully. 
A good idea is to save a flag when you successfully finished a job. To resume, you can first look at these flags to see which jobs have already been completed.

kennymckormick (approved):

PR #1191, opened 2021-09-28
[Improvement] Refactor Inference API

kennymckormick:
Motivation:
1. More flexible inference: The dataset type of the config and actual data format does not need to match
2. More modalities: Support skeleton-based
3. Flexible API: Can take path str, dictionary, or numpy array (to support torch serve) as inputs. 

@dreamerlin, @gengenkai  Please leave some comments if you have more ideas.

### DIFF
--- a/mmaction/apis/inference.py
+++ b/mmaction/apis/inference.py
@@ -5,6 +5,7 @@ import re
 from operator import itemgetter
 
 import mmcv
+import numpy as np
 import torch
 from mmcv.parallel import collate, scatter
 from mmcv.runner import load_checkpoint
@@ -37,13 +38,6 @@ def init_recognizer(config,
     elif not isinstance(config, mmcv.Config):
         raise TypeError('config must be a filename or Config object, '
                         f'but got {type(config)}')
-    if ((use_frames and config.dataset_type != 'RawframeDataset')
-            or (not use_frames and config.dataset_type != 'VideoDataset')):
-        input_type = 'rawframes' if use_frames else 'video'
-        raise RuntimeError('input data type should be consist with the '
-                           f'dataset type in config, but got input type '
-                           f"'{input_type}' and dataset type "
-                           f"'{config.dataset_type}'")
 
     # pretrained model is unnecessary since we directly load checkpoint later
     config.model.backbone.pretrained = None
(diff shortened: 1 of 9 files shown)


mmaction/apis/inference.py
```diff
@@ -81,30 +68,50 @@ def inference_recognizer(model,
...
+    if isinstance(video, dict):
+        input_flag = 'dict'
+    elif isinstance(video, np.ndarray):
+        assert len(video.shape) == 4, 'The shape should be T x H x W x C'
+        input_flag = 'array'
+    elif isinstance(video, str) and osp.exists(video):
+        if osp.isfile(video):
+            input_flag = 'video'
+        if osp.isdir(video):
+            input_flag = 'rawframes'
```
dreamerlin, line 81: what if the video is http url, I think we'd better to keep this function

mmaction/apis/inference.py
```diff
@@ -57,21 +51,14 @@ def init_recognizer(config,
...
 
 
-def inference_recognizer(model,
-                         video_path,
-                         label_path,
-                         use_frames=False,
-                         outputs=None,
-                         as_tensor=True):
+def inference_recognizer(model, video, outputs=None, as_tensor=True):
     """Inference a video with the detector.
```
dreamerlin, line 55: since ur here, please change this docstring -> recognizer

mmaction/apis/inference.py
```diff
@@ -14,10 +15,7 @@
 from mmaction.models import build_recognizer
 
 
-def init_recognizer(config,
-                    checkpoint=None,
-                    device='cuda:0',
-                    use_frames=False):
+def init_recognizer(config, checkpoint=None, device='cuda:0'):
```
dreamerlin, line 18: Better to use `kwargs` to get `use_frames` and provide deprecated infos

mmaction/apis/inference.py
```diff
@@ -57,21 +47,14 @@ def init_recognizer(config,
...
 
-def inference_recognizer(model,
-                         video_path,
-                         label_path,
-                         use_frames=False,
-                         outputs=None,
-                         as_tensor=True):
-    """Inference a video with the detector.
+def inference_recognizer(model, video, outputs=None, as_tensor=True):
+    """Inference a video with the recognizer.
```
dreamerlin, line 51: since the API is changed, better to use `kwargs` to get `label_path` and `use_frames` and provide deprecated infos

PR #1202, opened 2021-10-03
[Fix] switch the default value of persistent_workers to False

kennymckormick:
If the default value of persistent_workers is True and cannot be configured in the config file.  You can not train with num_workers=0.


dreamerlin (approved):

dreamerlin:
And `tools/misc/clip_feature_extraction.py`, `tools/test.py`

kennymckormick:
> And `tools/misc/clip_feature_extraction.py`, `tools/test.py`

Not applicable for these scripts: for inference, there is no need to  use persistent_workers, so it can always be set as OFF.

PR #1203, opened 2021-10-04
[Doc] Improve ERROR message

kennymckormick:


mmaction/datasets/builder.py
```diff
@@ -76,7 +76,7 @@ def build_dataloader(dataset,
             the worker processes after a dataset has been consumed once.
             This allows to maintain the workers Dataset instances alive.
             The argument also has effect in PyTorch>=1.8.0.
-            Default: True
+            Default: False
```
dreamerlin, line 79: Why change the default value to False, for PyTorch version?

kennymckormick: sorry, this PR is not checked out from the master branch, we can first deal with another PR

dreamerlin (approved):

PR #1227, opened 2021-10-20
[Minor] Rename: formating -> formatting

kennymckormick:


kennymckormick:
@dreamerlin , I think we do not need to add some deprecated warnings since, in all documentations, we told users to import components from `datasets.pipelines`.

PR #1306, opened 2021-12-03
[ModelZoo] Improve some sthv1 related models

dreamerlin:


kennymckormick:
We also need to: update the metafiles and update the links (We now use **config_name** as the filename of log and json)

kennymckormick (approved):

PR #1343, opened 2021-12-22
[Fix] Fix wrong variable name in ava_utils.py

jamiechoi1995:
The `action_id` in `read_csv()` should be `label`


kennymckormick:
Maybe I didn't quite get it. What's the purpose of this PR, I think action_id is just an internal variable (You can call it whatever you want. )

jamiechoi1995:
> Maybe I didn't quite get it. What's the purpose of this PR, I think action_id is just an internal variable (You can call it whatever you want. )

In my opinion, when someone needs to modify the code to fit their own dataset, using the wrong name can cause misunderstanding.

kennymckormick:
As mentioned here, we directly adapt this part from the [official AVA evaluation code](https://github.com/activitynet/ActivityNet/blob/master/Evaluation/get_ava_performance.py). So we prefer not to change them.

jamiechoi1995:
Got it. You can close this PR or leave it open. I am OK with it.

Issue #1382, opened 2022-01-12
TypeError: FastRCNN: AVARoIHead: LFBInferHead: __init__() got an unexpected keyword argument 'pretrained'


michael-camilleri:
## Description
I am trying to run Spatio-Temporal Action Localisation on a small sample of the AVA dataset (to get me started with the library), but am running into an issue with the LFB model. Specifically, the generation of the LFB features yields the following Exception:
"TypeError: FastRCNN: AVARoIHead: LFBInferHead: __init__() got an unexpected keyword argument 'pretrained'"

#### Installation
I installed MMCV, MMDet and MMAction2 from source and also from pip at different points to debug the error (each time tearing down the conda environment), but the issue remains.

This gives me the following envrionment:
```bash
sys.platform: linux
Python: 3.7.11 (default, Jul 27 2021, 14:32:16) [GCC 7.5.0]
CUDA available: True
GPU 0: Quadro K620
CUDA_HOME: /usr/local/cuda-10.2
NVCC: Cuda compilation tools, release 10.2, V10.2.89
GCC: gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
PyTorch: 1.10.1+cu102
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) Math Kernel Library Version 2020.0.0 Product Build 20191122 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.2.3 (Git Hash 7336ca9f055cf1bfa13efb658fe15dc9b41f0740)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 10.2
  - NVCC architecture flags: -gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_70,code=sm_70
  - CuDNN 7.6.5
  - Magma 2.5.2
[...]

PR #1386, opened 2022-01-14
Fixed Issue #1382 and a Printout Bug

michael-camilleri:
## Motivation
As per Issue #1382, Spatio-Temporal Localisation using the LFB model was failing.
In addition, there was a slight bug in the printing of the assertion error when the exclusion file had too many (or too little) columns.

## Modification
 1. Added a `pretrained` argument (default None, and unused) to the `__init__()` of LFBInferHead
 2. Refactored the way the length of a row is displayed in `ava_utils.read_exclusions()`


kennymckormick:
Great, thx a lot for ur contribution.

PR #1406, opened 2022-01-26
[Fix] Fix bug in nondistributed multi-gpu training

kennymckormick:


michael-camilleri:
If someone can approve this, I think it is a useful update. I have had issues with training failing at the end  (although I could recover all data) due to non-distributed training.

kennymckormick:
> If someone can approve this, I think it is a useful update. I have had issues with training failing at the end (although I could recover all data) due to non-distributed training.

Hi, Michael.  We merged this PR and added a message to encourage users to use distributed training instead.

PR #1414, opened 2022-02-02
[Improvement] Only support distributed training

kennymckormick:
Here is a suggestion: can we deprecate the option `launcher=None` for training and testing? Below are the reasons: 

1. In practice, we seldomly use `launcher=None`,  thus it takes additional efforts to validate this option when codes are updated. 
2. In most cases, you can achieve the same effect with `launcher=pytorch`. For example, you can set the number of GPUs to be used in `dist_train.sh`, or you can set which GPU to be used by setting `CUDA_VISIBLE_DEVICES` flags. 

Thus, due to Occam's razor principle, I think removing `launcher=None` is a good idea and it can save much effort for both users and maintainers. If get approved, I will further update the documentation.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -73,18 +73,6 @@ def parse_args():
         action='store_true',
         help=('whether to test the best checkpoint (if applicable) after '
               'training'))
-    group_gpus = parser.add_mutually_exclusive_group()
-    group_gpus.add_argument(
-        '--gpus',
-        type=int,
-        help='number of gpus to use '
-        '(only applicable to non-distributed training)')
-    group_gpus.add_argument(
-        '--gpu-ids',
-        type=int,
-        nargs='+',
-        help='ids of gpus to use '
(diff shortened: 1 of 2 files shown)


kennymckormick:
@ZwwWayne

tools/train.py
```diff
@@ -100,8 +88,8 @@ def parse_args():
         "'--cfg-options model.backbone.depth=18 model.backbone.with_cp=True'")
     parser.add_argument(
         '--launcher',
-        choices=['none', 'pytorch', 'slurm', 'mpi'],
```
ZwwWayne, line 103: It seems that this PR introduces a lot of deprecations, which should be avoided. 
1. 'mpi' launch should be kept because we need it when running on parrots in K8s.
2. --gpus and --gpu-ids should be kept for a while with deprecation warning, then we can deprecate them.

PR #1434, opened 2022-02-11
Singlelabel support

michael-camilleri:
## Motivation

Support Single-Label Training i.e. where only one class label should be output (as opposed to multi-label) for the AVA DataSet.

## Modification

I have changed the BBoxHeadAVA class to support single-label training: I also modified the bbox2result within transforms.py for the same reason.

I have also updated unit-tests to support the above.

## BC-breaking (Optional)

It should not: however, linting is failing (yapf and docformatter) and not sure what is the issue.

## Checklist

1. Pre-commit or other linting tools should be used to fix the potential lint issues. [DONE]
2. The modification should be covered by complete unit tests. If not, please add more unit tests to ensure the correctness. [DONE]
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMCls. <Not sure>
4. The documentation should be modified accordingly, like docstring or example tutorials. [DONE]


kennymckormick:
Thanks a lot for the contribution! I will review this PR and respond in 1 / 2 days.

mmaction/core/bbox/transforms.py
```diff
@@ -3,35 +3,58 @@
...
+    assert scores.shape[-1] > 1
 
-    # We only handle multilabel now
-    assert labels.shape[-1] > 1
+    # Robustly check for multi/single-label:
+    if not hasattr(thr, '__len__'):
+        multilabel = thr >= 0
+        thr = (thr,) * num_classes
+    else:
+        multilabel = False
```
kennymckormick, line 42: I think it should be `multilabel = True` here. Since the default setting for the original code is `multilabel = True`.

mmaction/models/heads/bbox_head.py
```diff
@@ -177,22 +199,29 @@ def loss(self,
...
-            loss = bce_loss(cls_score, labels, reduction='none')
+            # Compute loss
+            loss = loss_func(cls_score, labels, reduction='none')
             pt = torch.exp(-loss)
             F_loss = self.focal_alpha * (1 - pt)**self.focal_gamma * loss
             losses['loss_action_cls'] = torch.mean(F_loss)
 
-            recall_thr, prec_thr, recall_k, prec_k = self.multi_label_accuracy(
-                cls_score, labels, thr=0.5)
+            recall_thr, prec_thr, recall_k, prec_k = self.topk_accuracy(
```
kennymckormick, line 222: BTW, if we use the single label, what about directly reporting the normal Top-k accuracy instead (Top-k: if top k predictions match the ground-truth label)? In that situation, I think we don't need to report accuracy-recall.

kennymckormick:
Nice work~ Please check the two comments left.

mmaction/core/bbox/transforms.py
michael-camilleri: Actually, you are right: that was a bug in my logic. Rather, it should be True, because when we have multiple individual thresholds it explicitly should indicate that we are multi-label. Thanks. Fixed in the next commit.

mmaction/models/heads/bbox_head.py
michael-camilleri: I guess you are referring to computing just accuracy rather than prec/recall.

It does make sense, although:
 1. I wanted to retain continuity with the previous methods as far as possible (i.e. minimal changes) - doing so would require more changes throughout the library to handle a different metric to be reported
 2. There is still merit to having precision/recall metrics even in in single-label classification (see e.g. https://www.analyticsvidhya.com/blog/2020/12/accuracy-and-its-shortcomings-precision-recall-to-the-rescue/)

michael-camilleri:
I am still a bit concerned on why my linting is failing: any ideas on what I could be doing wrong?

kennymckormick:
> I am still a bit concerned on why my linting is failing: any ideas on what I could be doing wrong?

Don't worry, I'll help to fix this.

michael-camilleri:
Please hold off on merging: I am writing unit tests for the loss function, and there may be some bug in the way the recall is computed.

michael-camilleri:
Ok: Bugs have been fixed. Pull request can be reviewed for merging (subject to Lint issues)

kennymckormick:
> Ok: Bugs have been fixed.
> 
> The main issue is that Single-label classification requires Pytorch>=1.10, so that the CrossEntropyLoss supports multi-target. How should I deal with this?

A problem is: in most implementations, the labels for cross entropy loss need to be int64, not a float vector. I think there are two possible solutions: 1). When multi_label == False, we assume that there is one and only one element in the label vector is 1 (others are 0), so you can manually convert onehot vectors to integer labels. 2). You can implement the cross entropy loss urself, which takes onehot vectors as labels (I think it is also simple).

michael-camilleri:
I will try option 2: implement it myself. The reason is that even in single-label classification, one may wish to pass soft-labels for training (e.g. when one is uncertain which is the true label, or for label-smoothing).

I will try to get round to this in the coming days.

michael-camilleri:
All tests seem to have succeeded, apart from linting. If you let me know what is wrong with yapf/docformatter, I could have a look at fixing it.

dreamerlin:
> All tests seem to have succeeded, apart from linting. If you let me know what is wrong with yapf/docformatter, I could have a look at fixing it.

You can refer to https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md

michael-camilleri:
Even on my machine, I am unable to install it and have it run.

Is there another way?

michael-camilleri:
Can someone review this please?

kennymckormick:
Lint problems fixed, merged into the master.

PR #1574, opened 2022-04-11
[Feature] Add confusion matrix to allowed metrics in BaseDataset

rlleshi:
Just a tiny change adding confusion matrix as an allowed metric as well as a basic visualization.

### DIFF
--- a/mmaction/utils/misc.py
+++ b/mmaction/utils/misc.py
@@ -36,6 +36,6 @@ def visualize_confusion_matrix(confusion_matrix):
     import matplotlib.pyplot as plt
     import seaborn as sns
     plt.figure(figsize=(15, 10))
-    sns.set(font_scale=1.4)
+    sns.set(font_scale=1.5)
     sns.heatmap(confusion_matrix, annot=True, square=True, cbar=True)
     plt.show()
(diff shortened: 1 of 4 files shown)


mmaction/utils/misc.py
```diff
@@ -25,3 +25,17 @@ def get_thread_id():
...
+
+    Args:
+        confusion_matrix (np.array): the confusion matrix
+    """
+    import matplotlib.pyplot as plt
+    import seaborn as sns
+    plt.figure(figsize=(15, 10))
+    sns.set(font_scale=1.5)
+    sns.heatmap(confusion_matrix, annot=True, square=True, cbar=True)
+    plt.show()
```
kennymckormick, line 41: Better to save it as an image file to a specified location.

kennymckormick:
BTW, since confusion matrix is a matrix, we can not say one matrix is better than another. Try to add some assertions to make sure `confusion_matrix` is not used for `save_best` during validation.

rlleshi:
> BTW, since confusion matrix is a matrix, we can not say one matrix is better than another. Try to add some assertions to make sure `confusion_matrix` is not used for `save_best` during validation.

[This](https://github.com/open-mmlab/mmaction2/blob/660ec046039dc9a39d98e8bbbc6d260cbb7a9957/mmaction/core/evaluation/eval_hooks.py#L157) takes care of that, right?

PR #1652, opened 2022-06-01
[Fix] build_audio_features

rlleshi:
hotfix glob search

### DIFF
--- a/tools/data/build_audio_features.py
+++ b/tools/data/build_audio_features.py
@@ -297,9 +297,7 @@ if __name__ == '__main__':
 
     mmcv.mkdir_or_exist(args.spectrogram_save_path)
 
-    files = glob.glob(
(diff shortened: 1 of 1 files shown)


rlleshi:
Could someone merge this? It's just a single line.

ly015:
Thanks for your help! I made a tiny change to add back the '.' before the extension.

ly015 (approved):

rlleshi:
That was the original problem :sweat_smile: 

Otherwise, the default arg should be changed from `default='.m4a'` to just `m4a`

ly015:
I think the default arg is `m4a`: https://github.com/open-mmlab/mmaction2/blob/f189c04f4392c9acb79c596e6567a2ed296bc041/tools/data/build_audio_features.py#L293

Dai-Wenxun:
Fixed in #2570

Issue #1781, opened 2022-08-12
validation by iteration causes model.training=False to continue until the end of an epoch


makecent:
**Describe the bug**

During the validation, the state of `model.training` will be trun to `False`:
https://github.com/open-mmlab/mmaction2/blob/5fa8faa6a7cb2adf9b5eaf9f8a7473b65bb8228d/mmaction/apis/test.py#L76

While if validation by iteration, e.g., `evaluation = dict(by_epoch=False, interval=1000, metrics=['top_k_accuracy'])`, then the state `model.training=False` will continue [until the next epoch](https://github.com/open-mmlab/mmcv/blob/f527e43c1a1e52e5903410255aea1f71f005a161/mmcv/runner/epoch_based_runner.py#L43) when  `EpochBasedRunner` was adopted. This makes the remaining iterations in the epoch where the evaluation happens to be excuted under `eval` state.

hukkai:
Thanks @makecent ! So we just need to switch to model.train() before the end of this function? Could you submit a PR?

makecent:
> Thanks @makecent ! So we just need to switch to model.train() before the end of this function? Could you submit a PR?

I am not sure if this should be a kind of BUG of the `mmcv` because `IterBasedRunner` should work well with validation by iterations. 

Since mmaction2 only uses `EpochBasedRunner`, we may **either** 

- adopt `IterBasedRunner` when some conditions are met, e.g. `evalution.by_epoch=False`
- propose this as a BUG in an issue of `mmcv`.

The choice depends on how you `mmlab` expects the user to use the `Runner`: Should the evaluation by **iteration** be allowed when `EpochBasedRunner` is used?

hukkai:
In the next release mmlab repo series will use a new engine. Will discuss this with the related developers.

hukkai:
The new release is at dev-1.x branch. If you have any further questions, feel free to re-open the issue. Thanks!

PR #1784, opened 2022-08-13
Resolving the issue caused by the EvalHook changing the model training state to False

makecent:
This is just a demo PR for resolving issue #1781. 

When valiation by iterations, i.e., `evaluation.by_epoch=False`, the model training state will be trun off during the validation, i.e., `runner.model.training=False`. And the evaluation state will continue until the next epoch, causing the remaining iterations in current epoch to be wrong.

So I copied the `EpochBasedRunner` from the mmcv and add `model.train()` before `run_iter` to make sure model in the correct state in every iter.

This is PR just for description. To apply the modified `EpochBasedRunner`, some import issues need to be ground. Instead of this, I prefer merging this PR to the `mmcv`.

### DIFF
--- a/mmaction/core/runner/epoch_based_runner.py
+++ b/mmaction/core/runner/epoch_based_runner.py
@@ -1,0 +1,199 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import os.path as osp
+import platform
+import shutil
+import time
+import warnings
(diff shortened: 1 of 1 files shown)


hukkai:
Transfered to `mmcv` https://github.com/open-mmlab/mmcv/issues/2199

hukkai:
Closed since this will be discussed in `mmcv`

PR #2206, opened 2023-02-08
[Enhance] support auto import 

cir7:
## Motivation
[feat] support auto import


requirements/mminstall.txt
```diff
@@ -1,2 +1,2 @@
 mmcv>=2.0.0rc0,<2.1.0
-mmengine>=0.3.0
```
ly015, line 2: Please also update `mmaction/__init__.py`

ly015: Consider updating to mmengine v0.5.0?

ly015 (approved):

PR #2480, opened 2023-05-17
[Fix] Fix a bug of MViT when set with_cls_token to False 

KeepLost:
Resolve the problem when class token is taken away in MViT. 

Put more precisely, when `with_cls_token` and `output_cls_token` are both set `False` in model `MViT`, an error would occur because its `MultiScaleBlock`and `MultiScaleAttention` components don't pass the `with_cls_token` argument from class `MViT`, and set it `True` by default.


cir7:
Thanks for your contribution! this is indeed a bug, we will merge it after review.

cir7 (approved):

PR #2561, opened 2023-06-22
Update build_rawframes.py for multiple GPU support

HoBeom:
## Motivation
https://github.com/open-mmlab/denseflow/issues/19#issuecomment-1253316054
Updated code for multiple GPU support as the num_gpu option is unused and left unattended confusing.
it aims to detect and use multiple GPUs automatically.

## Modification

Changing the unused num_gpu option.
Use the pid to assign the GPU to each process.


HoBeom:
I think it would be appropriate to merge this commit into the master branch.

Issue #2630, opened 2023-08-08
[Bug] FormatShape bug for optical flow


makecent:
### Branch

main branch (1.x version, such as `v1.0.0`, or `dev-1.x` branch)

### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmaction2/issues) and [Discussions](https://github.com/open-mmlab/mmaction2/discussions) but cannot get the expected help.
- [X] I have read the [documentation](https://mmaction2.readthedocs.io/en/latest/) but cannot get the expected help.
- [X] The bug has not been fixed in the [latest version](https://github.com/open-mmlab/mmaction2).

### Environment

N/A

### Describe the bug

As the **Normalization operation** now is moved from the pipeline to the **data preprocessor**. The necessary processing step for optical flows, which is originally in the `Normalize` pipeline in the older version, is missing:

In **older version**, the `Normalize` pipeline **stacks** the `flow_x` and `flow_y`:

https://github.com/open-mmlab/mmaction2/blob/02d5d9b84a66914a69fc2e569029c4fe1b81ebf3/mmaction/datasets/pipelines/augmentations.py#L1207-L1232

In `1.x` version, the stacking operation is lost as the `Normalize` pipeline is no longer used. Causing the dimension error in `FormatShape`:
```terminal
  File "/home/louis/miniconda3/envs/mmengine/lib/python3.8/site-packages/mmaction/datasets/transforms/formatting.py", line 260, in transform
    imgs = np.transpose(imgs, (0, 1, 5, 2, 3, 4))
  File "<__array_function__ internals>", line 180, in transpose
  File "/home/louis/.local/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 660, in transpose
    return _wrapfunc(a, 'transpose', axes)
  File "/home/louis/.local/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 57, in _wrapfunc
    return bound(*args, **kwds)
ValueError: axes don't match array
```

Dai-Wenxun:
For the formatting of the optical flow, you need to set the `input_format ` of `FormatShape ` as `NCHW_Flow`, as shown [here](https://github.com/open-mmlab/mmaction2/blob/main/mmaction/datasets/transforms/formatting.py#L307). The code of this branch will stack the optical flow at the last dimension as in your #2631.

Dai-Wenxun:
For the normalization of optical flow, I think we can implement it as follows: 

```python

[...]

PR #2631, opened 2023-08-08
[Enhance] Support 2D&3D Optical Flow Training

makecent:
## Motivation

Fixing the error when decoding optical flows as now the `Normalize` pipeline is no longer used. See issue #2630

## Modification

Modify the RawFrameDecode transform to stack `x_frame` and `y_frame`, making the variable `imgs` of optical flow consistent with RGB, albeit in channel dimension 2 vs 3.


Dai-Wenxun:
LGTM

Dai-Wenxun (approved):

PR #2690, opened 2023-09-20
[Fix] Fix `gt_labels`

Dai-Wenxun:
Fix `gt_labels`


hukkai (approved):

stronggerlu:
I would like to ask if the modified ActionClip does not support training?

